PR #10, opened 2023-09-22
fix: use cache first when get StorageTrie/Trie

krish-nr:
TBD


core/state/database.go
```diff
@@ -238,19 +238,57 @@ func (db *cachingDB) purgeLoop() {
...
+			newTriesArray = [3]*triePair{
+				{root: root, trie: tr.ResetCopy()},
+				tries[0],
+				tries[1],
+			}
+		} else {
+			newTriesArray = [3]*triePair{{root: root, trie: tr.ResetCopy()}, nil, nil}
+		}
+		db.storageTrieCache.Add(addrHash, newTriesArray)
+	}
```
keroro520, line 291:
```suggestion

	db.CacheStorage(addrHash, root, tr)
```

How about reusing `db.CacheStorage()` to update the cache?

core/state/database.go
```diff
@@ -238,19 +238,57 @@ func (db *cachingDB) purgeLoop() {
...
+
 	tr, err := trie.NewStateTrie(trie.StateTrieID(root), db.triedb)
 	if err != nil {
 		return nil, err
 	}
+
+	//update LRU
+	if db.accountTrieCache != nil {
+		db.accountTrieCache.Add(root, tr.ResetCopy())
+	}
```
keroro520, line 256:
```suggestion

	db.CacheAccount(root, tr)
```

How about re-using `db.CacheAccount()` to update the cache?

bendanzhentan, line 256:
```suggestion

	db.CacheAccount(root, tr)
```

How about re-using `db.CacheAccount()` to update the cache?

core/state/database.go
bendanzhentan, line 291:
```suggestion

	db.CacheStorage(addrHash, root, tr)
```

How about reusing `db.CacheStorage()` to update the cache?

bendanzhentan (approved):
LGTM

core/state/database.go
```diff
@@ -238,19 +238,45 @@ func (db *cachingDB) purgeLoop() {
...
+		}
+	}
+
 	tr, err := trie.NewStateTrie(trie.StateTrieID(root), db.triedb)
 	if err != nil {
 		return nil, err
 	}
+
+	//update cache
+	db.CacheAccount(root, tr)
```
j75689, line 254: no need to cache here, it will be cached when `statedb.commit`

core/state/database.go
```diff
@@ -238,19 +238,45 @@ func (db *cachingDB) purgeLoop() {
...
+		}
+	}
+
 	tr, err := trie.NewStateTrie(trie.StorageTrieID(stateRoot, addrHash, root), db.triedb)
 	if err != nil {
 		return nil, err
 	}
+
+	//update cache
+	db.CacheStorage(addrHash, root, tr)
```
j75689, line 278: no needs to cache here, it will be cached when `statedb.commit`

core/state/database.go
```diff
@@ -238,19 +238,45 @@ func (db *cachingDB) purgeLoop() {
...
 	return tr, nil
 }
 
 // OpenStorageTrie opens the storage trie of an account.
 func (db *cachingDB) OpenStorageTrie(stateRoot common.Hash, addrHash, root common.Hash) (Trie, error) {
+	// try cache first
+	if db.storageTrieCache != nil {
+		if tries, exist := db.storageTrieCache.Get(addrHash); exist {
+			for _, triePair := range tries {
+				if triePair.root == root {
```
j75689, line 265:
```
if triePair != nil && triePair.root == root {
     return triePair.trie.(*trie.SecureTrie).Copy(), nil
}
```

core/state/database.go
bendanzhentan: As far as I understand, `StateDB.Commit()` only commits dirty modified entries; if an entry isn't modified, it won't be cached by `StateDB.Commit()`.

Therefore, I think it makes sense to explicitly cache entries here.

core/state/database.go
```diff
@@ -238,6 +238,13 @@ func (db *cachingDB) purgeLoop() {
 
 // OpenTrie opens the main account trie at a specific root hash.
 func (db *cachingDB) OpenTrie(root common.Hash) (Trie, error) {
+	//try cache first
+	if db.accountTrieCache != nil {
+		if tr, ok := db.accountTrieCache.Get(root); ok {
+			return tr.(*trie.SecureTrie), nil
```
boolafish, line 244: Shall we use `.Copy()` here instead?

The BSC implementation has the `Copy()`: https://github.com/bnb-chain/bsc/pull/257/files#diff-2002ca9f189ebd4dbe9ec777405505525bb4a67db44536243e615f66d85426d2R191

2129zxl: exactly, updated, thx

krish-nr: exactly, updated, thx

bendanzhentan (approved):

joeylichang:
Have you ever tested it in PBSS mode?

PR #11, opened 2023-09-22
fix: pass a SnapshotOption func when init a new pruner

krish-nr:
**Description**
NewPruner() creates a new snapshot, but it doesn't pass a SnapshotOption function, so the snapshot will always have a limit of 128 layers here.

**Rationale**
Add SnapshotOption config when creates a new snapshot
 
**Changes**
Pass SnapshotOption in NewPruner() when creating new snapshot

### DIFF
--- a/cmd/geth/snapshot.go
+++ b/cmd/geth/snapshot.go
@@ -23,6 +23,8 @@ import (
 	"os"
 	"time"
 
+	cli "github.com/urfave/cli/v2"
+
 	"github.com/ethereum/go-ethereum/cmd/utils"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/rawdb"
@@ -35,7 +37,6 @@ import (
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/ethereum/go-ethereum/trie"
-	cli "github.com/urfave/cli/v2"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


cmd/geth/snapshot.go
```diff
@@ -173,8 +174,14 @@ func pruneState(ctx *cli.Context) error {
...
 		BloomSize: ctx.Uint64(utils.BloomFilterSizeFlag.Name),
 	}
-	pruner, err := pruner.NewPruner(chaindb, prunerconfig,
-		pruner.WithTriesInMemory(ctx.Uint64(utils.TriesInMemoryFlag.Name)))
+	pruner, err := pruner.NewPruner(chaindb, prunerconfig, pruner.CombinedOptions{
+		PrunerOptions: []pruner.PrunerOption{
+			pruner.WithTriesInMemory(ctx.Uint64(utils.TriesInMemoryFlag.Name)),
+		},
+		SnapshotOptions: []snapshot.SnapshotOption{
+			snapshot.SetCapLimit(int(ctx.Uint64(utils.TriesInMemoryFlag.Name))),
```
keroro520, line 182: Please help confirm that it is intended to reuse `utils.TriesInMemoryFlag` as `snapshot.capLimit`?

bendanzhentan, line 182: Please help confirm that it is intended to reuse `utils.TriesInMemoryFlag` as `snapshot.capLimit`?

2129zxl: Brilliant point!
All occurrences of 'capLimit' in the current code are set to the same value as 'TriesInMemory'. However, based on the current mechanism, this doesn't seem necessary. Therefore, we need to decide whether to introduce an additional separate configuration option specifically for 'capLimit' after confirming some details.

bendanzhentan (approved):

j75689 (approved):

PR #23, opened 2023-10-18
fix: allow a previously underpriced transaction back in after a timeout 

andyzhang2023:
This issue had been reported to official repo in go-ethereum, here is it [#28018](https://github.com/ethereum/go-ethereum/issues/28018).
And the fix solution had been adopted by official team, with the merged pr [#28097](https://github.com/ethereum/go-ethereum/issues/28018).

### DIFF
--- a/eth/fetcher/tx_fetcher.go
+++ b/eth/fetcher/tx_fetcher.go
@@ -54,8 +54,9 @@ const (
 	// re-request them.
 	maxTxUnderpricedSetSize = 32768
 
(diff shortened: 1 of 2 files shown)


bendanzhentan (commented):
Is it possible to introduce these updates via cherry-pick? It keeps the source author when use cherry-pick.

andyzhang2023:
> Is it possible to introduce these updates via cherry-pick? It keeps the source author when use cherry-pick.

Yes, i redo it with cherry-pick here: https://github.com/bnb-chain/op-geth/pull/34,

PR #33, opened 2023-11-22
feat: add option to reannounce remote transactions

andyzhang2023:
### Description
This new feature allow us to re-announce on remote transactions, but not only local transactions.

### Rationale
The re-announce only applied on local transaction so far, but in some special network architectures, transactions need to be passed to miner via P2P network, which means the re-announce don't work on them. That's the reason why we add this option.

### Example
```
--txpool.reannounceremotes=true
```
Its default value is false.

### Changes

Notable changes:
* remote transactions can be re-announce as well as local ones.


owen-reorg (approved):

redhdx:
Is re-announce on local transactions enabled by default? Can it be set?

redhdx (approved):

owen-reorg:
> Is re-announce on local transactions enabled by default? Can it be set?

yes, check this: https://github.com/bnb-chain/op-geth/pull/2

PR #35, opened 2023-11-29
feat: Drop non-local transactions whose gas fee cap is below or equal to the potential minimum base fee

owen-reorg:
### Description

Drop non-local transactions whose gas fee cap is below or equal to the potential minimum base fee

### Rationale

Currently, there are transactions with a gas fee cap lower than 8 wei (the potential minimum base fee according to the current EIP1559 configuration). These transactions will never be included in a block. As a result, they remain in the transaction pool until it is full and they are kicked out by transactions with a higher gas price.

This pull request aims to prevent such transactions from entering the transaction pool.

### Example

NA

### Changes

Notable changes:
* add an additional rule in txpool


Rakibibul (commented):
L7ke

PR #39, opened 2023-12-12
frequently clean staled pending transaction 

andyzhang2023:
add expiration on pending transaction

### Description

Staled pending transactions is cleaned frequently, to save the memory for other transactions. 

### Rationale

As a high-throughput chain, transactions with too low gas price might occupy too much memory of the transaction pool, so we decide to clean the staled (exists in pool for over 3 hours) pending transactions, in a frequency of once /1min.

### Changes
Notable changes:
* a cleaning crontab job is added for staled pending transaction


andyzhang2023:
need further discussion.

PR #40, opened 2023-12-17
improve txpool reannounce frequency

andyzhang2023:
### Description

### Rationale

### Example

### Changes

Notable changes:
* ...


andyzhang2023:
used for debugging and should be closed

PR #45, opened 2024-01-03
feat: add TrieCommitInterval configuration, commit trie every TrieCommitInterval blocks.

welkin22:
### Description

Currently, all of our nodes can only be set to archive mode for gcMode. As the amount of data gradually increases, each of our nodes needs to be able to configure to full mode. This mode will delete Trie data that is lower than the latest block height minus TriesInMemory blocks, instead of committing and storing it in the database. This way, our nodes will occupy less disk space and have better performance. However, our official bridge and proposer components both need to use Trie data from about 1 hour ago. After enabling full mode, both the official bridge and proposer components will not function properly.
We need to enable full mode, and at the same time, we need to solve the problem of the official bridge and proposer using Trie data.

### Rationale

It should be noted that the full mode does not discard all Trie data. It has a gcproc statistic that tracks the cumulative time spent processing blocks. When gcproc is greater than flushInterval (default is 1 hour), the full mode will commit and store the complete Trie data for the current block height minus TriesInMemory  in the database. On the other hand, the submission frequency of our Mainnet and Testnet proposers is fixed and controlled by the configuration l2OutputOracleSubmissionInterval. We can use this configuration to calculate which block heights the official bridge and proposers need to use. Therefore, we add a TrieCommitInterval configuration in geth, whose value is equal to the value of l2OutputOracleSubmissionInterval for the corresponding network. In our code, we add a judgment condition that when the target block height (i.e. the latest block height minus TriesInMemory) divided by TrieCommitInterval results in a remainder of 0, it means that the target block height is the block height where we need to save the Trie data. We then commit the Trie data and store it in the database.

### Example
add TrieCommitInterval config in config.toml
```toml
[Eth]
TrieCommitInterval = 240
```
then add `gcMode=full` to geth param.
All Trie data for block heights that are multiples of 240 will be available.

### Changes

Notable changes:
* Add TrieCommitInterval configuration option
* Add an additional condition after bc.gcproc > flushInterval, where chosen block height is divisible by TrieCommitInterval, commit triedb.

### DIFF
--- a/eth/ethconfig/gen_config.go
+++ b/eth/ethconfig/gen_config.go
@@ -3,6 +3,7 @@
 package ethconfig
 
 import (
+	"math/big"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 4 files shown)


krish-nr (approved):
/LGTM

eth/ethconfig/gen_config.go
```diff
@@ -295,5 +323,26 @@ func (c *Config) UnmarshalTOML(unmarshal func(interface{}) error) error {
...
+	}
+	if dec.RollupHistoricalRPC != nil {
+		c.RollupHistoricalRPC = *dec.RollupHistoricalRPC
+	}
+	if dec.RollupHistoricalRPCTimeout != nil {
+		c.RollupHistoricalRPCTimeout = *dec.RollupHistoricalRPCTimeout
+	}
+	if dec.RollupDisableTxPoolGossip != nil {
+		c.RollupDisableTxPoolGossip = *dec.RollupDisableTxPoolGossip
+	}
```
owen-reorg, line 346: Why this configs change? Seems not related to the change.

owen-reorg (approved):

PR #49, opened 2024-01-19
fix: ignore errors that caused by gap to keep peer connection  

krish-nr:
### Description

This PR is designed to address scenarios where dynamic-type peers in geth are too easily removed. In certain scenarios, we can increase the tolerance.

### Rationale

In the context of opbnb, our block production rate and the number of transactions per block mean that there is a gap between the time geth finishes writing and the node's broadcast time. During this interval, for clients making requests to that node, it is possible that geth has not yet completed the persistence of the requested data at the moment of the request. This can trigger a peer disconnected event, even though the peer's status is actually healthy. Subsequently, extra waiting time is spent before the peer is added back. We would like peers in such situations to maintain the connection. When the block height is not more than a certain threshold behind, the connection should not be disconnected.

### Example

N/A

### Changes

Notable changes:
* define and add err types
* ignore specific err in handle loop


redhdx (approved):

welkin22 (approved):
LGTM

owen-reorg (approved):

PR #56, opened 2024-01-26
fix: use triesInMemory in downloader

welkin22:
### Description

Currently, our downloader has a hardcoded value of fsMinFullBlocks set to 64, which is half of 128. Since 128 represents TriesInMemory, which is now a configurable option in our geth version, the value of fsMinFullBlocks should change accordingly. For example, if we set TriesInMemory to 32, then the value of fsMinFullBlocks should be 16. We need to resolve this issue.

### Rationale

We will store the values of the TriesInMemory configuration item and the fsMinFullBlocks field in the Downloader, and then replace all previous references to the hardcoded fsMinFullBlocks.

### Example

none

### Changes

Notable changes:
* Added TriesInMemory and fsMinFullBlocks field to Downloader.
* Replace all references to hard-coded fsMinFullBlocks


redhdx (dismissed):

bendanzhentan (dismissed):

welkin22:
Due to https://github.com/bnb-chain/op-geth/pull/58 modifying the relevant logic, this PR needs to wait for that PR to be merged before deciding whether to merge.

welkin22:
Close, as we no longer need the triesInMemory parameter.

PR #63, opened 2024-02-20
Fix prefetcher concurrent map read and write issue

andyzhang2023:
### Description
It's a bug. The triePrefetcher structure should be thread safe, and it occasionally crashes the service with fatal error of "concurrent map read and write" on the map `fetchers`. We just add a lock to protected the map here.

### Changes
 the map `fetchers` will be protected by an RWMutex, and there are five function calls in which the map `fetchers` will be accessed:
* prefetch(): read and write, need to be Lock
* trie(): read only, need to be RLock
* copy(): read only, need to be RLock
* close(): read and wirte, need to be Lock
* used(): read only, need to be RLock


core/state/trie_prefetcher.go
```diff
@@ -190,7 +225,11 @@ func (p *triePrefetcher) trie(owner common.Hash, root common.Hash) Trie {
 // used marks a batch of state items used to allow creating statistics as to
 // how useful or wasteful the prefetcher is.
 func (p *triePrefetcher) used(owner common.Hash, root common.Hash, used [][]byte) {
-	if fetcher := p.fetchers[p.trieID(owner, root)]; fetcher != nil {
+	p.fetchersLock.Lock()
```
welkin22, line 228: The purpose of fetchersLock is to protect the fetchers map from concurrent read and write operations. In this case, fetchers is only being read and not written, so perhaps an RLock is sufficient.
As for the change `fetcher.used = used`, it should not be within the scope of the lock's protection.

andyzhang2023: It sounds good! Applied.

welkin22 (approved):
LGTM

owen-reorg (approved):
LGTM

owen-reorg:
close this issue for now since we removed the feature in the v0.3.0 release

PR #64, opened 2024-02-21
fix(op-geth):fix static peer cannot reassign header tasks

krish-nr:
### Description

disconnect and reconnect peer when a header gap occurs

### Rationale

when ignore header anchor, static peers cannot reassign header tasks with the same connection, reconnect need to be processed to get reassigned header link task

### Example

N/A

### Changes

Notable changes:
* disconnect and reconnect peer when a header gap error occurs


welkin22 (approved):
LGTM

bendanzhentan (approved):
Discussed with the author.
The main problem here is that the errorous connection can't be reassigned header tasks. This PR serves as a temporary solution.

PR #65, opened 2024-02-23
Fix/recover txpool nolocals configuration

andyzhang2023:
### Description

This change adds a check to ensure that transactions added to the legacy pool are not treated as 'locals' if the global locals-management has been disabled.


welkin22 (approved):
LGTM

PR #66, opened 2024-03-05
chore: add more metrics for perf

will-2012:
### Description

Add more metrics for perf.

### Rationale

N/A.

### Example

`--metrics.expensive` enable more metrics for perf.

### Changes

Notable changes:
* N/A.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1882,12 +1882,10 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool) (int, error)
 		storageUpdateTimer.Update(statedb.StorageUpdates)               // Storage updates are complete(in validation)
 		accountHashTimer.Update(statedb.AccountHashes)                  // Account hashes are complete(in validation)
 		storageHashTimer.Update(statedb.StorageHashes)                  // Storage hashes are complete(in validation)
-		triehash := statedb.AccountHashes + statedb.StorageHashes       // The time spent on tries hashing
-		trieUpdate := statedb.AccountUpdates + statedb.StorageUpdates   // The time spent on tries update
 		trieRead := statedb.SnapshotAccountReads + statedb.AccountReads // The time spent on account read
 		trieRead += statedb.SnapshotStorageReads + statedb.StorageReads // The time spent on storage read
-		blockExecutionTimer.Update(ptime - trieRead)                    // The time spent on EVM processing
-		blockValidationTimer.Update(vtime - (triehash + trieUpdate))    // The time spent on block validation
+		blockExecutionTimer.Update(ptime)                               // The time spent on EVM processing
+		blockValidationTimer.Update(vtime)                              // The time spent on block validation
 
 		// Write the block to the chain and get the status.
 		var (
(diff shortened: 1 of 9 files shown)


core/blockchain.go
```diff
@@ -1882,12 +1882,10 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool) (int, error)
...
 		accountHashTimer.Update(statedb.AccountHashes)                  // Account hashes are complete(in validation)
 		storageHashTimer.Update(statedb.StorageHashes)                  // Storage hashes are complete(in validation)
-		triehash := statedb.AccountHashes + statedb.StorageHashes       // The time spent on tries hashing
-		trieUpdate := statedb.AccountUpdates + statedb.StorageUpdates   // The time spent on tries update
 		trieRead := statedb.SnapshotAccountReads + statedb.AccountReads // The time spent on account read
 		trieRead += statedb.SnapshotStorageReads + statedb.StorageReads // The time spent on storage read
-		blockExecutionTimer.Update(ptime - trieRead)                    // The time spent on EVM processing
-		blockValidationTimer.Update(vtime - (triehash + trieUpdate))    // The time spent on block validation
+		blockExecutionTimer.Update(ptime)                               // The time spent on EVM processing
+		blockValidationTimer.Update(vtime)                              // The time spent on block validation
```
bendanzhentan, line 1888: It seems that the definitions of `blockExecutionTimer` and `blockValidationTimer` have been re-defined. While I am not sure if this is intended. if yes, could you please explain why this change is better?

core/rawdb/accessors_snapshot.go
```diff
@@ -74,6 +76,10 @@ func DeleteSnapshotRoot(db ethdb.KeyValueWriter) {
 
 // ReadAccountSnapshot retrieves the snapshot entry of an account trie leaf.
 func ReadAccountSnapshot(db ethdb.KeyValueReader, hash common.Hash) []byte {
+	start := time.Now()
+	if metrics.EnabledExpensive {
+		defer func() { rawdbGetAccountSnapNodeTimer.UpdateSince(start) }()
+	}
```
bendanzhentan, line 82:
```suggestion
	if metrics.EnabledExpensive {
		start := time.Now()
		defer func() { rawdbGetAccountSnapNodeTimer.UpdateSince(start) }()
	}
```

core/state_processor.go
```diff
@@ -99,6 +106,9 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 		}
 		receipts = append(receipts, receipt)
 		allLogs = append(allLogs, receipt.Logs...)
+		if metrics.EnabledExpensive {
+			processTxTimer.UpdateSince(start)
+		}
```
bendanzhentan, line 111: IMO we should call `processTxTimer.UpdateSince(start)` even if `applyTransaction()` fails.

core/rawdb/accessors_snapshot.go
will-2012: Updated.

core/state_processor.go
will-2012: Fixed.

core/blockchain.go
will-2012: Yes; this PR does not delete any metrics, but only adjusts the calculation method of metrics, because when analyzing performance bottlenecks during `--metrics.expensive`, blockExecutionTimer/blockValidationTimer/blockWriteTimer may be negative when deducting concurrency latency. This is a bit confusing. In fact, if we pay attention to the metrics of internal concurrent actions, we can directly configure a corresponding panel, and  I think it is better to have clear semantics of blockExecutionTimer/blockValidationTimer/blockWriteTimer.   :)
This is just a slight change and ultimately depends on the reviewers' suggestions.

will-2012: Without `--metrics.expensive`, These `triehash/trieUpdate/..Commits` are all equal to 0.

bendanzhentan (approved):
LGTM

core/blockchain.go
bendanzhentan: ACK. Thanks.

owen-reorg:
Close this PR since most of the metrics have been added in https://github.com/bnb-chain/op-geth/pull/71

PR #76, opened 2024-03-27
bugfix: fix Resubscribe deadlock when unsubscribing after inner sub ends

welkin22:
### Description

See https://github.com/ethereum-optimism/optimism/issues/8086
We encountered the same issue, where a deadlock caused op-node to be unable to gracefully shutdown.
Therefore, we need to introduce this fix as soon as possible to resolve our current op-node issue.

### Rationale

`op-node` uses the `ResubscribeErr` method, but the size of the unsubscribed channel is 0, causing a deadlock during the shutdown process.

### Example

none

### Changes

Notable changes:
* fix `ResubscribeErr` method issue
* add ut


owen-reorg (approved):

andyzhang2023 (approved):
Seams good.

PR #77, opened 2024-03-28
feat: EVM execution opcode level optimization

sunny2022da:
several optimizations included:
- opcode fusion
- jumpdest calculation optimization
- SHA caching in interpreter

we observed ~1.2-1.5x speedup in execution time of opbnb mainnet when syncing block at hight ~11m

### Description

This PR introduce several optimizations in EVM execution engine, mainly 3 points:
opcode fusion: Combine several hottest opcodes into one super instruction to reduce strength
jumpdest calculation optimizations:  reduce the unnecessary jumpdest calculation every time a call happened by leverage caching. 
SHA cacheing in interpreter:  optimized the SHA opcode to use hash cache that already implemented in geth. 

### Rationale

To speed up the execution time of contracts and hence enhance the performance.

### Example
N/A

### Changes

Notable changes:
* opcode fusion
* jumpdest calculation optimization
* SHA caching in interpreter


sunny2022da:
UT test fails has been fixed. The current 2 failed cases seems not related with this PR.
- TestPrecompiledPointEvaluation: seems related with ZK
- TestExecutionSpec: seems related with gas but it fails too without this PR.

FYI

core/vm/evm.go
```diff
@@ -294,7 +304,11 @@ func (evm *EVM) CallCode(caller ContractRef, addr common.Address, input []byte,
 		// Initialise a new contract and set the code that is to be used by the EVM.
 		// The contract is a scoped environment for this execution context only.
 		contract := NewContract(caller, AccountRef(caller.Address()), value, gas)
-		contract.SetCallCode(&addrCopy, evm.StateDB.GetCodeHash(addrCopy), evm.StateDB.GetCode(addrCopy))
+		// try get from cache
+		code := evm.StateDB.GetCode(addr)
```
owen-reorg, line 308:
```suggestion
		code := evm.StateDB.GetCode(addrCopy)
```

Though the effect might be the same, I think it's better to align the previous pattern.

core/vm/errors.go
```diff
@@ -37,7 +37,6 @@ var (
 	ErrGasUintOverflow          = errors.New("gas uint64 overflow")
 	ErrInvalidCode              = errors.New("invalid code: must not begin with 0xef")
 	ErrNonceUintOverflow        = errors.New("nonce uint64 overflow")
-
```
owen-reorg, line 40: Let's avoid the unnecessary changes to avoid further merge conflicts.

core/vm/gas_table_test.go
```diff
@@ -149,7 +149,6 @@ func TestCreateGas(t *testing.T) {
 			if tt.eip3860 {
 				config.ExtraEips = []int{3860}
 			}
-
```
owen-reorg, line 152: Let's avoid the unnecessary changes to avoid further merge conflicts.

core/vm/instructions.go
```diff
@@ -237,11 +241,13 @@ func opKeccak256(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) (
 
 	if interpreter.hasher == nil {
 		interpreter.hasher = crypto.NewKeccakState()
-	} else {
+	} /* else {
 		interpreter.hasher.Reset()
-	}
-	interpreter.hasher.Write(data)
-	interpreter.hasher.Read(interpreter.hasherBuf[:])
+	}*/
```
owen-reorg, line 246: Shall we rewrite this part and remove the comments?

crypto/crypto.go
```diff
@@ -112,7 +112,8 @@ func CreateAddress(b common.Address, nonce uint64) common.Address {
 // CreateAddress2 creates an ethereum address given the address bytes, initial
 // contract code hash and a salt.
 func CreateAddress2(b common.Address, salt [32]byte, inithash []byte) common.Address {
-	return common.BytesToAddress(Keccak256([]byte{0xff}, b.Bytes(), salt[:], inithash)[12:])
+	addr := common.BytesToAddress(Keccak256([]byte{0xff}, b.Bytes(), salt[:], inithash)[12:])
+	return addr
```
owen-reorg, line 116: seems no effect, can be revert

cmd/utils/flags.go
```diff
@@ -1050,6 +1051,13 @@ Please note that --` + MetricsHTTPFlag.Name + ` must be set to start the server.
 		Value:    metrics.DefaultConfig.InfluxDBOrganization,
 		Category: flags.MetricsCategory,
 	}
+
+	VMOpcodeOptimizeFlag = &cli.BoolFlag{
+		Name:     "vm.opcode.optimize",
+		Usage:    "enable opcode optimization",
+		Value:    true,
```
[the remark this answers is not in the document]
sunny2022da: our test show that the 'value: true' here takes no effect, if the argument is not set manually, the optimization is disabled by default.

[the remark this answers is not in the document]
sunny2022da: but I will delete the "value: true" here in case of misunderstanding

cmd/utils/flags.go
```diff
@@ -1901,6 +1909,11 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 		cfg.EnablePreimageRecording = ctx.Bool(VMEnableDebugFlag.Name)
 	}
 
+	if ctx.IsSet(VMOpcodeOptimizeFlag.Name) {
+		cfg.EnableOpcodeOptimizing = ctx.Bool(VMOpcodeOptimizeFlag.Name)
+		compiler.EnableOptimization()
```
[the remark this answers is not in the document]
sunny2022da: nice catch! will make the change.

core/vm/analysis.go
```diff
@@ -76,6 +76,14 @@ func codeBitmapInternal(code, bits bitvec) bitvec {
...
 		op := OpCode(code[pc])
 		pc++
+
+		// handle super instruction.
+		step, processed := codeBitmapForSI(code, pc, op, &bits)
+		if processed {
+			pc += step
+			continue
+		}
+
```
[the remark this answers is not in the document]
sunny2022da: it is ok because when optimization is disabled, the codebitmap here does't encounter any super instruction in this method, so the method takes no effect. On the otherside, it is hard in this go file to get the VMConfig and hence hard to add logic to know whether the optimization is disabled. So I suggest we leave it there.

owen-reorg (approved):

PR #81, opened 2024-04-10
feat: introduce hash cache to improve performance

sunny2022da:
This is part of the optimizations from BSC, microbench show the change could introduce  ~20% speed up of simple erc20 transfer scenario.

The original PR of BSC: https://github.com/bnb-chain/bsc/pull/257

### DIFF
--- a/crypto/crypto.go
+++ b/crypto/crypto.go
@@ -24,6 +24,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"github.com/VictoriaMetrics/fastcache"
 	"hash"
 	"io"
 	"math/big"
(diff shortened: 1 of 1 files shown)


crypto/crypto.go
```diff
@@ -24,6 +24,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"github.com/VictoriaMetrics/fastcache"
```
rickyyangz, line 27: format the imports

crypto/crypto.go
```diff
@@ -66,31 +69,52 @@ func NewKeccakState() KeccakState {
...
 	kh.Reset()
 	kh.Write(data)
 	kh.Read(h[:])
+	keccakState256Cache.Set(data, h.Bytes())
 	return h
 }
 
 // Keccak256 calculates and returns the Keccak256 hash of the input data.
 func Keccak256(data ...[]byte) []byte {
+	if len(data) == 1 {
```
rickyyangz, line 84: could you add some comments why we need some special logic for length == 1

sunny2022da: Just confirmed with the original contributor, the `length==1` case is the most common case which means that the method only accept one byte array as input. 
And by processing on the one input case, it is also reduce the complexity of the logic here since there is no need to concern the concatenation of bytearrays and hash cache calculations.

sunny2022da: I will add brief comments in the code.

owen-reorg (approved):

welkin22 (approved):
LGTM

PR #85, opened 2024-04-16
improve Pending() of txpool to reduce the latency when miner worker committing transactions

andyzhang2023:
### Description

improve Pending() of txpool to reduce the latency when the miner worker commits transactions, by maintaining a copy of pending transactions for the worker, to avoid lock race condition.

### Rationale
The miner worker has limited time to commit transactions, and it wastes too much to call Pending() to fetch transactions from the mem pool, which sometimes costs hundreds of milliseconds.

### Changes

Notable changes:
* Pending() method of txpool returns a copy of pending transactions instead of the original pending list.


redhdx (dismissed):
LGTM

core/txpool/legacypool/cache_for_miner.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func newCacheForMiner() *cacheForMiner {
+	return &cacheForMiner{
+		pending: make(map[common.Address]map[*types.Transaction]struct{}),
+		locals:  make(map[common.Address]bool),
+	}
+}
+
+func (pc *cacheForMiner) add(txs types.Transactions, signer types.Signer) {
```
owen-reorg, line 34: Will pass by pointer(`txs *types.Transactions`) be more efficient?

andyzhang2023: The `types.Transactions` is defined as ```type Transactions []*Transaction ```, so it doesn't take too much memory and should be ok to use `types.Transactions` directly.

owen-reorg (approved):

redhdx (approved):
LGTM

PR #89, opened 2024-04-22
feat(txpool): improve performance of Reheap

andyzhang2023:
### Description
Reheap() needs to be finished as quickly as possible to support a high-throughput pool. 

### Rationale

By profiling the CPU usages, we find that the GasTipCap() and GasFeeCap() cost the most, so we rewrite the compare function to avoid the non-sense calling of new(big.Int) and big.Set. 

cpuprofiling


redhdx (approved):
LGTM

owen-reorg (approved):

PR #112, opened 2024-05-21
feature(op-geth): add precompile for secp256r1 curve after haber hardfork

redhdx:
### Description

Add precompile for secp256r1 curve after haber hardfork.

### Rationale

The PR is about EIP-7212 which proposes the addition of a new precompiled contract to the EVM that allows signature verifications in the “secp256r1” elliptic curve by given parameters of message hash, r - s components of the signature, and x - y coordinates of the public key.

### Example

Add new precompiled contract after haber hardfork.

### Changes

Notable changes:
* Add new precompiled contract after haber hardfork.


owen-reorg (approved):

krish-nr (approved):
LGTM

PR #121, opened 2024-06-13
pathdb: handle persistent id when using nodebufferlist

sysvm:
### Description

When using nodebufferlist in pathdb, geth should maintain three node cache in memory for withdrawal proof. This pr fixes a case that when the length of history is greater than `StateHistory` but not beyond maxBufferSize, nodebufferlist still should keep three node cache.

Start a load test that uses 1000 accounts and 1TPS for about 50 hours to verify this pr, metrics are shown below:
image

### Rationale

N/A

### Example

N/A

### Changes

Notable changes:
* N/A

### DIFF
--- a/trie/triedb/pathdb/nodebufferlist.go
+++ b/trie/triedb/pathdb/nodebufferlist.go
@@ -270,6 +270,9 @@ func (nf *nodebufferlist) flush(db ethdb.KeyValueStore, clean *fastcache.Cache, 
 	}
 
 	commitFunc := func(buffer *multiDifflayer) bool {
+		if nf.count <= nf.rsevMdNum {
+			return false
+		}
 		if err := nf.base.commit(buffer.root, buffer.id, buffer.block, buffer.layers, buffer.nodes); err != nil {
 			log.Crit("failed to commit nodes to base node buffer", "error", err)
 		}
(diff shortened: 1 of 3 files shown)


trie/triedb/pathdb/nodebufferlist.go
```diff
@@ -270,12 +270,18 @@ func (nf *nodebufferlist) flush(db ethdb.KeyValueStore, clean *fastcache.Cache,
 	}
 
 	commitFunc := func(buffer *multiDifflayer) bool {
+		if nf.count <= nf.rsevMdNum {
+			log.Info("Skip forcing flush buffer list due to buffer list is too less",
```
joeylichang, line 274: I think this log is not accurate， maybe ("keep %d mulitidlfflayer in node bufferlist for getting withdrawal proof", nf.rsevMdNum) is better.

joeylichang (dismissed):

trie/triedb/pathdb/nodebufferlist.go
sysvm: fixed

joeylichang (dismissed):

joeylichang (approved):

krish-nr (approved):
LGTM

PR #128, opened 2024-07-12
enlarge the buffer size of transactions broadcaster

andyzhang2023:
### Description

enlarge the buffer size of transactions broadcaster from 4096 to 8192, to support higher TPS throughput of P2P broadcasting.


krish-nr (approved):
LGTM

PR #130, opened 2024-07-18
feature(op-geth): add opbnb gasless solution

redhdx:
### Description

Add opbnb gasless solution. This is a breaking change for opbnb. In order to enable opbnb to support gasless transactions, this PR introduces the new hard fork `wright`. After enabling the `wright` hard fork, opbnb's basefee will be 0. For transactions with a gasprice of 0, opbnb will not charge L1 cost. With this change, opbnb has added bundle transaction and bundle pool. Please refer https://github.com/bnb-chain/BEPs/pull/322 to the specific implementation details.

### Rationale

Add opbnb gasless solution which can enable opbnb to support gasless transaction.

### Changes

Notable changes:
* add new hard fork `wright`.
* add new apis about bundle.
  - eth_bundlePrice: returns the price of a bundle
  - eth_simulateGaslessBundle: simulates a gasless bundle
  - eth_sendBundle: sends a bundle
* add bundle pool.
* building block add mev logic.
* add new config.
  - --mev.enable: enable mev feature, can enable bundle pool and receive bundle transactions
  - --mev.bundle.receiver.url: enable bundle deliver function, can deliver bundle to the specific url.
  - --mev.bundle.gasprice.floor: set bundle gas price floor, default value is 1.
  - --bundlepool.globalslots: set bundle pool size, default value is 4096.


core/txpool/bundlepool/bundlepool.go
```diff
@@ -0,0 +1,443 @@
...
+		select {
+		case bundle := <-p.deliverBundleCh:
+			for url, cli := range p.sequencerClients {
+				cli := cli
+				url := url
+				go func() {
+					var hash common.Hash
+					err := cli.CallContext(context.Background(), &hash, "eth_sendBundle", *bundle)
+					if err != nil {
+						log.Error("failed to deliver bundle to sequencer", "url", url, "err", err)
```
welkin22, line 128: Maybe it's better to add some metrics. We can know in time that a bundle failed to be passed to the sequencer, what is the failure rate, etc.

core/txpool/bundlepool/bundlepool.go
```diff
@@ -0,0 +1,443 @@
...
+			return
+		}
+	}
+}
+
+func (p *BundlePool) register(url string) error {
+	var sequencerClient *rpc.Client
+	if url != "" {
+		var err error
+		sequencerClient, err = rpc.Dial(url)
```
welkin22, line 143: Suggest adding some retries to the dialing process, you can refer to the `dialRPCClientWithBackoff` method used in op-node

miner/worker_builder.go
```diff
@@ -0,0 +1,415 @@
...
+
+	start := time.Now()
+	pending := w.eth.TxPool().Pending(true)
+	packFromTxpoolTimer.UpdateSince(start)
+	log.Debug("packFromTxpoolTimer", "duration", common.PrettyDuration(time.Since(start)), "hash", env.header.Hash())
+
+	// Split the pending transactions into locals and remotes.
+	localTxs, remoteTxs := make(map[common.Address][]*txpool.LazyTransaction), pending
+
+	// TODO will remove after fix txpool perf issue
```
welkin22, line 65: pls remove unnecessary comments

core/txpool/bundlepool/bundlepool.go
```diff
@@ -0,0 +1,445 @@
...
+	price, err := p.simulator.SimulateBundle(bundle)
+	if err != nil {
+		return err
+	}
+
+	if p.mevConfig.SentryEnabled {
+		p.deliverBundleCh <- originBundle
+	}
+
+	if price.Cmp(p.minimalBundleGasPrice()) < 0 && p.slots+numSlots(bundle) > p.config.GlobalSlots {
```
welkin22, line 191: This filter condition is after line 188, which causes some transactions that do not meet the price conditions and exceed the slots to be propagated to the sequencer. Is this expected behavior?

cmd/utils/flags.go
```diff
@@ -558,6 +558,28 @@ var (
...
+	}
+	SentryEnabledFlag = &cli.BoolFlag{
+		Name:     "sentry.enable",
+		Usage:    "Enable sentry",
+		Category: flags.MinerCategory,
+	}
+	SequencerUrl = &cli.StringFlag{
+		Name:     "sequencer.url",
+		Usage:    "Url of sequencer endpoint to use. Multiple urls are supported, separated by commas",
+		Value:    "http://127.0.0.1:8545",
```
owen-reorg, line 574: This default value is nonsense. I think we can remove it.

cmd/utils/flags.go
```diff
@@ -1718,6 +1740,25 @@ func setMiner(ctx *cli.Context, cfg *miner.Config) {
...
+		cfg.Mev.MevEnabled = ctx.Bool(MevEnabledFlag.Name)
+	}
+	if ctx.IsSet(SentryEnabledFlag.Name) {
+		cfg.Mev.SentryEnabled = ctx.Bool(SentryEnabledFlag.Name)
+	}
+	if ctx.IsSet(SequencerUrl.Name) {
+		url := ctx.String(SequencerUrl.Name)
+		if strings.Contains(url, ",") {
+			cfg.Mev.Sequencers = strings.Split(url, ",")
+		} else {
```
owen-reorg, line 1753: There is no need to add `if else` here. `strings.Split(url, ",")` works for both branch.

cmd/utils/flags.go
```diff
@@ -558,6 +558,28 @@ var (
 		Value:    ethconfig.Defaults.Miner.NewPayloadTimeout,
 		Category: flags.MinerCategory,
 	}
+	MevEnabledFlag = &cli.BoolFlag{
```
owen-reorg, line 561: We should make the flag names and description more clear.

1. Add a universal prefix, e.g. `mev`.
2. `sequencer.url` is not accurate, the forward url is not necessary to be a sequencer.
3. pricefloor is better to be bundle gas price floor.
4. `mev.enable` is better to be `mev.miner.enable`.

core/txpool/bundlepool/config.go
```diff
@@ -0,0 +1,73 @@
...
+type Config struct {
+	PriceLimit uint64 // Minimum gas price to enforce for acceptance into the pool
+	PriceBump  uint64 // Minimum price bump percentage to replace an already existing transaction (nonce)
+
+	GlobalSlots     uint64 // Maximum number of bundle slots for all accounts
+	GlobalQueue     uint64 // Maximum number of non-executable bundle slots for all accounts
+	MaxBundleBlocks uint64 // Maximum number of blocks for calculating MinimalBundleGasPrice
+
+	BundleGasPricePercentile      uint8         // Percentile of the recent minimal mev gas price
+	BundleGasPricerExpireTime     time.Duration // Store time duration amount of recent mev gas price
```
owen-reorg, line 18: plz refine the comments to match the words(mev gas price -> bundle gas price)

core/txpool/bundlepool/config.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	if conf.MaxBundleBlocks < 1 {
+		log.Warn("Sanitizing invalid txpool max bundle blocks", "provided", conf.MaxBundleBlocks, "updated", DefaultConfig.MaxBundleBlocks)
+		conf.MaxBundleBlocks = DefaultConfig.MaxBundleBlocks
+	}
+	if conf.BundleGasPricePercentile >= 100 {
+		log.Warn("Sanitizing invalid txpool bundle gas price percentile", "provided", conf.BundleGasPricePercentile, "updated", DefaultConfig.BundleGasPricePercentile)
+		conf.BundleGasPricePercentile = DefaultConfig.BundleGasPricePercentile
+	}
+	if conf.BundleGasPricerExpireTime < 1 {
```
owen-reorg, line 64: 1 means 1 Nanosecond here, does it make sense?

miner/worker.go
```diff
@@ -819,6 +831,30 @@ func (w *worker) commitTransaction(env *environment, tx *types.Transaction) ([]*
...
+		return nil, err
+	}
+	if receipt.Status == types.ReceiptStatusFailed && unRevertible {
+		return nil, errors.New("no revertible transaction failed")
+	}
+	env.txs = append(env.txs, tx)
+	env.receipts = append(env.receipts, receipt)
+	if w.config.Mev.MevEnabled {
+		minerFee, _ := tx.EffectiveGasTip(env.header.BaseFee)
+		env.profit.Add(env.profit, new(big.Int).Mul(new(big.Int).SetUint64(receipt.GasUsed), minerFee))
```
owen-reorg, line 856: Is it possible to also consider the L1 fee in the profit?

miner/worker.go
```diff
@@ -1231,10 +1293,15 @@ func (w *worker) generateWork(genParams *generateParams) *newPayloadResult {
 	innerExecutionTimer.Update(core.DebugInnerExecutionDuration)
 
 	log.Debug("build payload statedb metrics", "parentHash", genParams.parentHash, "accountReads", common.PrettyDuration(work.state.AccountReads), "storageReads", common.PrettyDuration(work.state.StorageReads), "snapshotAccountReads", common.PrettyDuration(work.state.SnapshotAccountReads), "snapshotStorageReads", common.PrettyDuration(work.state.SnapshotStorageReads), "accountUpdates", common.PrettyDuration(work.state.AccountUpdates), "storageUpdates", common.PrettyDuration(work.state.StorageUpdates), "accountHashes", common.PrettyDuration(work.state.AccountHashes), "storageHashes", common.PrettyDuration(work.state.StorageHashes))
-
+	fees := big.NewInt(0)
+	if w.config.Mev.MevEnabled {
+		fees = work.profit
```
owen-reorg, line 1298:
```
	if w.config.Mev.MevEnabled {
		minerFee, _ := tx.EffectiveGasTip(env.header.BaseFee)
		env.profit.Add(env.profit, new(big.Int).Mul(new(big.Int).SetUint64(receipt.GasUsed), minerFee))
	}
```

What's the difference of the profit and totalFees?

core/txpool/bundlepool/bundlepool.go
```diff
@@ -0,0 +1,456 @@
...
+}
+
+func (p *BundlePool) deliverLoop() {
+	for {
+		select {
+		case bundle := <-p.deliverBundleCh:
+			for url, cli := range p.bundleReceiverClients {
+				cli := cli
+				url := url
+				go func() {
```
owen-reorg, line 129: I didn't get the design here. If we are going to send the bundles in goroutines anyway, why do we bother use `deliverBundleCh`?
Why don't we just have a `deliverBundle` function and use a separate goroutine to call it when we want to deliver the bundle?

If we want to control the concurrency, we should use a buffered channel and a reasonable number of consumers to do the delivery.

miner/worker_builder.go
```diff
@@ -0,0 +1,406 @@
...
+	for _, tx := range txs {
+		if interrupt != nil {
+			if signal := interrupt.Load(); signal != commitInterruptNone {
+				return signalToErr(signal)
+			}
+		}
+		// If we don't have enough gas for any further transactions then we're done.
+		if env.gasPool.Gas() < params.TxGas {
+			log.Trace("Not enough gas for further transactions", "have", env.gasPool, "want", params.TxGas)
+			break
```
owen-reorg, line 105: Will this break happen in the middle of a bundle, which results in partial txns of a bundle included in block?

miner/worker_builder.go
```diff
@@ -0,0 +1,406 @@
...
+	if err != nil {
+		log.Error("fail to simulate bundles base on the same state", "err", err)
+		return nil, nil, err
+	}
+
+	// sort bundles according to fresh gas price
+	sort.SliceStable(simulatedBundles, func(i, j int) bool {
+		priceI, priceJ := simulatedBundles[i].BundleGasPrice, simulatedBundles[j].BundleGasPrice
+
+		return priceI.Cmp(priceJ) >= 0
```
owen-reorg, line 198: Use `>=` in stable sort will make it unstable. Should use `>` instead.

Check this example: https://go.dev/play/p/wwyLf916mrD.

owen-reorg: BTW, according to the doc https://pkg.go.dev/sort#SliceStable, it's recommend to use https://pkg.go.dev/slices#SortStableFunc instead.

core/txpool/bundlepool/bundlepool.go
redhdx: Simply implemented the retry logic, please reveiw again.

miner/worker_builder.go
redhdx: Removed. please review again.

miner/worker_builder.go
```diff
@@ -0,0 +1,415 @@
...
+			// Strange error, discard the transaction and get the next in line (note, the
+			// nonce-too-high clause will prevent us from executing in vain).
+			log.Error("Transaction failed in the bundle", "hash", tx.Hash(), "err", err)
+			return signalToErr(commitInterruptBundleCommit)
+		}
+	}
+
+	if !w.isRunning() && len(coalescedLogs) > 0 {
+		// We don't push the pendingLogsEvent while we are mining. The reason is that
+		// when we are mining, the worker will regenerate a mining block every 3 seconds.
```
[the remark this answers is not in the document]
redhdx: fixed, please review again.

core/txpool/bundlepool/bundlepool.go
redhdx: No, it is unreasonable logic. We should deliver bundle to receiver after these conditions. Fixed, please review again.

cmd/utils/flags.go
redhdx: Removed, please review again.

cmd/utils/flags.go
redhdx: Fixed, please review again.

cmd/utils/flags.go
redhdx: Yes, make sense. I think we just need `mev.enalbe` flag to enable bundlepool, `mev.bundle.receiver.url` flag to decide whether to deliver bundle and `mev.bundle.gasprice.floor` to be bundle gas price floor. Please review again.

core/txpool/bundlepool/config.go
redhdx: Removed useless configuration, please review again.

core/txpool/bundlepool/config.go
redhdx: Removed useless configuration and check, please review again.

core/types/bundle.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/rlp"
+)
+
+const (
+	// MaxBundleAliveBlock is the max alive block for bundle
+	MaxBundleAliveBlock = 300
+	// MaxBundleAliveTime is the max alive time for bundle
+	MaxBundleAliveTime = 5 * 60 // second
```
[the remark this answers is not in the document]
redhdx: I think this is a two-dimensional configuration. If we still keep two parameters in the bundle structure, it is better to keep it consistent with bsc in bundle structure, and although our block time is one second, the block time may be changed one day in the future.

eth/api_backend.go
```diff
@@ -323,6 +324,31 @@ func (b *EthAPIBackend) SendTx(ctx context.Context, signedTx *types.Transaction)
...
+
+	if len(bundles) == 0 {
+		return gasFloor
+	}
+
+	sort.SliceStable(bundles, func(i, j int) bool {
+		return bundles[j].Price.Cmp(bundles[i].Price) < 0
+	})
+
+	idx := len(bundles) / 2
```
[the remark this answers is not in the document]
redhdx: In fact, I have removed most of the useless configurations. For this, I think taking the median gasprice as return value is reasonable. I don’t expect it to be changed anymore. The second thing is that I think if this value is configurable , it is the node behavior, it will make the return price logic unstable and confusing.

miner/worker.go
redhdx: I think our L1 fee is not used as profit, but to cover the fees we pay when submitting transactions to L1. If L1 fee is taken into consideration, it is not easy for us to simply add it directly, because we also need to consider the actual payment of L1.

miner/worker.go
redhdx: The calculation formulas of the profit and totalFees are the same. Totalfees is a loop to calculate all transactions, which is a relatively time-consuming operation. We now have two workers working, so my idea is to calculate it during the concurrent execution of each go routine. Calculate the result instead of waiting and then start two go routines to calculate the totalfees of the two workers. At the same time, regarding the atomicity of the bundle, set the profit to 0 to discard the worker. It is relatively simple to implement. Finally, we return the result if there is profit can be used directly without calculating totalfees again.

miner/worker.go
redhdx: I probably misunderstood what you meant. It should mean that when adding the each transaction fee, for a transaction with a gas price of 0, the fee of the transaction is deducted by l1 fee and then added as profit, right?

redhdx: If it it, I think it is reasonable and feasible, but there may be a problem here. The profit we ultimately consider here is the entire worker, which may include multiple bundles. There is the possibility that some bundles may be paid for by other bundles. , the bundle gas price floor is still needed.

miner/worker.go
redhdx: Maybe something needs to be different here, totalfees is just to calculate miner fees, I think our profit needs to consider basefee and l1 fee, because basefee is our income and l1 fee is our cost.

ethclient/ethclient.go
```diff
@@ -667,6 +667,23 @@ func toCallArg(msg ethereum.CallMsg) interface{} {
...
+	if err != nil {
+		return common.Hash{}, err
+	}
+	return hash, nil
+}
+
+// BundlePrice returns the price of a bundle
+func (ec *Client) BundlePrice(ctx context.Context) *big.Int {
+	var price *big.Int
+	_ = ec.c.CallContext(ctx, &price, "eth_bundlePrice")
```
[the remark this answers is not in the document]
redhdx: fixed, please review again.

internal/ethapi/api_bundle.go
```diff
@@ -0,0 +1,118 @@
...
+
+	currentHeader := s.b.CurrentHeader()
+
+	if args.MaxBlockNumber == 0 && (args.MaxTimestamp == nil || *args.MaxTimestamp == 0) {
+		maxTimeStamp := currentHeader.Time + types.MaxBundleAliveTime
+		args.MaxTimestamp = &maxTimeStamp
+	}
+
+	if args.MaxBlockNumber != 0 && args.MaxBlockNumber > currentHeader.Number.Uint64()+types.MaxBundleAliveBlock {
+		return common.Hash{}, newBundleError(errors.New("the maxBlockNumber should not be lager than currentBlockNum + 300"))
```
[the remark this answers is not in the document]
redhdx: fixed, please review again.

core/txpool/bundlepool/bundlepool.go
redhdx: There is indeed unnecessary to use channel to implement, which increases complexity. Now it is simplified, please review again.

miner/worker_builder.go
redhdx: Yes, it is incorrect. Fixed, please review again.

miner/worker_builder.go
redhdx: Refer doc, refine it, please review again.

eth/ethconfig/config.go
```diff
@@ -170,8 +171,9 @@ type Config struct {
 	Miner miner.Config
 
 	// Transaction pool options
-	TxPool   legacypool.Config
-	BlobPool blobpool.Config
+	TxPool     legacypool.Config
+	BlobPool   blobpool.Config
+	BundlePool bundlepool.Config
```
welkin22, line 176: The gen_config.go file needs to be regenerated

redhdx: fixed, please review again.

core/txpool/bundlepool/bundlepool.go
redhdx: Added, please help review again.

core/txpool/bundlepool/bundlepool.go
```diff
@@ -0,0 +1,443 @@
...
+			if err != nil {
+				bundleDeliverFailed.Inc(1)
+				log.Error("failed to deliver bundle to receiver", "url", url, "err", err)
+			}
+		}()
+		bundleDeliverAll.Inc(1)
+	}
+
+	for p.slots+numSlots(bundle) > p.config.GlobalSlots {
+		p.drop()
```
owen-reorg, line 200: I think every time when we dropped a bundle, we should compare the price again. Otherwise, a huge bundle whose price is only higher the the minimal price bundle will result in other bundles dropped.

core/txpool/bundlepool/bundlepool.go
```diff
@@ -0,0 +1,455 @@
...
+
+const (
+	// TODO: decide on a good default value
+	// bundleSlotSize is used to calculate how many data slots a single bundle
+	// takes up based on its size. The slots are used as DoS protection, ensuring
+	// that validating a new bundle remains a constant operation (in reality
+	// O(maxslots), where max slots are 4 currently).
+	bundleSlotSize = 128 * 1024 // 128KB
+
+	maxMinTimestampFromNow = int64(300) // 5 minutes
```
owen-reorg, line 33: Should this align with max alive time 60?

miner/worker_builder.go
```diff
@@ -0,0 +1,450 @@
...
+		BundleGasFees:  new(big.Int),
+		BundleGasUsed:  0,
+		BundleGasPrice: new(big.Int),
+	}
+
+	for _, bundle := range bundles {
+		prevState := currentState.Copy()
+		prevGasPool := new(core.GasPool).AddGas(gasPool.Gas())
+
+		// the floor gas price is 99/100 what was simulated at the top of the block
```
owen-reorg, line 277: why is 99/100?

welkin22 (approved):
LGTM

miner/worker_builder.go
```diff
@@ -0,0 +1,450 @@
...
+			"txcount", len(simulatedBundle.OriginalBundle.Txs))
+
+		includedTxs = append(includedTxs, bundle.OriginalBundle.Txs...)
+
+		mergedBundle.BundleGasFees.Add(mergedBundle.BundleGasFees, simulatedBundle.BundleGasFees)
+		mergedBundle.BundleGasUsed += simulatedBundle.BundleGasUsed
+
+		for _, tx := range includedTxs {
+			if !containsHash(bundle.OriginalBundle.RevertingTxHashes, tx.Hash()) {
+				env.UnRevertible.Add(tx.Hash())
```
owen-reorg, line 307: if this happen, the bundle should be dropped, but I didn't find this logic.

owen-reorg:
plz add testnet fork time: 8.15 2pm utc+8

owen-reorg (approved):

welkin22 (approved):

miner/worker.go
```diff
@@ -822,6 +832,22 @@ func (w *worker) commitTransaction(env *environment, tx *types.Transaction) ([]*
 	return receipt.Logs, nil
 }
 
+func (w *worker) commitBundleTransaction(env *environment, tx *types.Transaction, unRevertible bool) ([]*types.Log, error) {
+	if tx.Type() == types.BlobTxType {
+		return w.commitBlobTransaction(env, tx)
```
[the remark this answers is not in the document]
redhdx: Refer commitTransaction function.Reject blob tx in API request.

miner/worker_builder.go
redhdx: Use this logic In commitBundleTransaction, drop bundle: `receipt.status == ReceiptStatusFailed && unRevertible`.

redhdx:
> plz add testnet fork time: 8.15 2pm utc+8

Added. Plz confirm it.

PR #131, opened 2024-07-24
fix: ignore truncation target range as unclean shutdown happens

krish-nr:
### Description

ignore truncation target range as unclean shutdown happens

### Rationale

bufferlist meet an update failure when unclean shutdown happens and lead to a panic, ignore the range judgement in this scenario

### Example

N/A

### DIFF
--- a/trie/triedb/pathdb/disklayer.go
+++ b/trie/triedb/pathdb/disklayer.go
@@ -331,14 +331,20 @@ func (dl *diskLayer) commit(bottom *diffLayer, force bool) (*diskLayer, error) {
 	if overflow {
 		if _, ok := dl.buffer.(*nodebufferlist); ok {
 			persistentID := rawdb.ReadPersistentStateID(dl.db.diskdb)
-			if persistentID > limit {
-				oldest = persistentID - limit + 1
-				log.Info("Forcing prune ancient under nodebufferlist", "disk_persistent_state_id",
-					persistentID, "truncate_tail", oldest)
-			} else {
-				log.Info("No prune ancient under nodebufferlist, less than db config state history limit")
+			if limit >= persistentID {
+				log.Info("No prune ancient under nodebufferlist, less than db config state history limit", "persistent_id", persistentID, "limit", limit)
 				return ndl, nil
 			}
+			targetOldest := persistentID - limit + 1
+			realOldest, err := dl.db.freezer.Tail()
+			if err == nil && targetOldest <= realOldest {
+				log.Info("No prune ancient under nodebufferlist due to truncate oldest less than real oldest, which maybe happened in abnormal restart",
+					"tartget_oldest_id", targetOldest, "real_oldest_id", realOldest, "error", err)
+				return ndl, nil
+			}
+			oldest = targetOldest
+			log.Info("Forcing prune ancient under nodebufferlist", "disk_persistent_state_id",
+				persistentID, "truncate_tail", oldest)
 		}
 
 		pruned, err := truncateFromTail(ndl.db.diskdb, ndl.db.freezer, oldest-1)
--- a/trie/triedb/pathdb/disklayer.go
+++ b/trie/triedb/pathdb/disklayer.go
@@ -332,7 +332,7 @@ func (dl *diskLayer) commit(bottom *diffLayer, force bool) (*diskLayer, error) {
 		if _, ok := dl.buffer.(*nodebufferlist); ok {
 			persistentID := rawdb.ReadPersistentStateID(dl.db.diskdb)
 			if limit >= persistentID {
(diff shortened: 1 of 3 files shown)


will-2012:
https://github.com/bnb-chain/op-geth/blob/887404faf9cb5d5c0cf3b3a974889a71d20436f2/trie/triedb/pathdb/disklayer.go#L332-L342

==>
```
                if _, ok := dl.buffer.(*nodebufferlist); ok {
			persistentID := rawdb.ReadPersistentStateID(dl.db.diskdb)
			if limit >= persistentID {
				log.Info("No prune ancient under nodebufferlist, less than db config state history limit", "persistent_id", persistentID, "limit", limit)
				return ndl, nil
			}
			targetOldest := persistentID - limit + 1
			realOldest, err := dl.db.freezer.Tail()
			if err == nil && targetOldest <= realOldest {
				log.Info("No prune ancient under nodebufferlist due to truncate oldest less than real oldest, which maybe happened in abnormal restart",
					"tartget_oldest_id", targetOldest, "real_oldest_id", realOldest, "error", err)
				return ndl, nil
			}
			oldest = targetOldest
			log.Info("Forcing prune ancient under nodebufferlist", "disk_persistent_state_id",
				persistentID, "truncate_tail", oldest)
		}
```
maybe better.

The problem probably occurs because there may be a gap between write wal and write stateid. Write wal occurs during commit to disklayer, and write stateid occurs during disklayer background flush. Therefore, stateid (ReadPersistentStateID) may be smaller than the actual wal head, and stateid - limit may be smaller than the actual wal tail.

trie/triedb/pathdb/history.go
```diff
@@ -619,7 +619,8 @@ func truncateFromTail(db ethdb.Batcher, freezer *rawdb.ResettableFreezer, ntail
 	}
 	// Ensure that the truncation target falls within the specified range.
 	if otail > ntail || ntail > ohead {
-		return 0, fmt.Errorf("out of range, tail: %d, head: %d, target: %d", otail, ohead, ntail)
+		log.Warn("truncate from tail out of range", "tail:", otail, "head:", ohead, "target:", ntail)
+		return 0, nil
```
will-2012, line 623: This interface can remain unchanged.

will-2012 (dismissed):

trie/triedb/pathdb/history.go
2129zxl: Exactly.
Theoretically, adding a conditional check in the outer layer can prevent out-of-bounds issues. However, setting the logic for repeated deletions as a critical error and directly causing a panic would be too strict. Therefore, we have also increased the tolerance at the interface level. Ideally, it should not enter this logic. We have added a warning log to capture such cases to see if there are any other anomalies, while ensuring that the program does not panic.

krish-nr: Exactly.
Theoretically, adding a conditional check in the outer layer can prevent out-of-bounds issues. However, setting the logic for repeated deletions as a critical error and directly causing a panic would be too strict. Therefore, we have also increased the tolerance at the interface level. Ideally, it should not enter this logic. We have added a warning log to capture such cases to see if there are any other anomalies, while ensuring that the program does not panic.

trie/triedb/pathdb/disklayer.go
```diff
@@ -331,14 +331,20 @@ func (dl *diskLayer) commit(bottom *diffLayer, force bool) (*diskLayer, error) {
...
 		if _, ok := dl.buffer.(*nodebufferlist); ok {
 			persistentID := rawdb.ReadPersistentStateID(dl.db.diskdb)
-			if persistentID > limit {
-				oldest = persistentID - limit + 1
-				log.Info("Forcing prune ancient under nodebufferlist", "disk_persistent_state_id",
-					persistentID, "truncate_tail", oldest)
-			} else {
-				log.Info("No prune ancient under nodebufferlist, less than db config state history limit")
+			if limit >= persistentID {
+				log.Info("No prune ancient under nodebufferlist, less than db config state history limit", "persistent_id", persistentID, "limit", limit)
```
sysvm, line 335: Maybe can lower `INFO` level log to `DEBUG` due to high frequency.

trie/triedb/pathdb/history.go
will-2012: Ok, got it～

trie/triedb/pathdb/disklayer.go
krish-nr: updated

will-2012 (approved):

joeylichang (approved):

will-2012:
Pls refine PR title. `as unclean shutdown happens` is misleading.

PR #132, opened 2024-07-31
(op-geth) Add extra error info when meet an unexpected el sync

krish-nr:
### Description

When op-geth exits due to certain exceptional circumstances and the truncate operation results in the deletion of head data, the block-by-block insertion requests from op-node will trigger an EL sync in geth. This PR addresses this situation.
related op-node pr: [https://github.com/bnb-chain/opbnb/pull/222](url)

### Rationale

Instead of ignoring the unexpected EL sync triggers, specific errors are returned. op-node will handle them based on the state.

### Example

N/A

### DIFF
--- a/eth/catalyst/api.go
+++ b/eth/catalyst/api.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


eth/catalyst/api.go
```diff
@@ -681,6 +682,9 @@ func (api *ConsensusAPI) delayPayloadImport(block *types.Block) (engine.PayloadS
 		// that the parent state is missing and the syncer rejected extending the
 		// current cycle with the new payload.
 		log.Warn("Ignoring payload with missing parent", "number", block.NumberU64(), "hash", block.Hash(), "parent", block.ParentHash(), "reason", err)
+		if strings.Contains(err.Error(), "forced head needed for startup") {
```
welkin22, line 685: Let's turn "forced head needed for startup" into a constant, so we can track where this error is being created and where it is being used.

krish-nr: updated

welkin22 (approved):
LGTM

redhdx (approved):
LGTM

PR #136, opened 2024-08-12
fix bundle pool issue

redhdx:
### Description

Fix bundle pool issue.

### Rationale

Every time when dropped a bundle, should compare the price again. Otherwise, a huge bundle whose price is only higher the the minimal price bundle will result in other bundles dropped.

### Example

N/A

### Changes

Notable changes:
* change bundle pool drop logic.

### DIFF
--- a/core/txpool/bundlepool/bundlepool.go
+++ b/core/txpool/bundlepool/bundlepool.go
@@ -200,7 +200,7 @@ func (p *BundlePool) AddBundle(bundle *types.Bundle, originBundle *types.SendBun
 	}
 
 	for p.slots+numSlots(bundle) > p.config.GlobalSlots {
-		p.drop()
+		p.drop(bundle)
 	}
 	p.bundles[hash] = bundle
 	heap.Push(&p.bundleHeap, bundle)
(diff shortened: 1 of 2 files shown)


owen-reorg:
plz add UT for this feature.

core/txpool/bundlepool/bundlepool.go
```diff
@@ -175,8 +175,10 @@ func (p *BundlePool) AddBundle(bundle *types.Bundle, originBundle *types.SendBun
 	p.mu.Lock()
 	defer p.mu.Unlock()
 
-	if price.Cmp(p.minimalBundleGasPrice()) <= 0 && p.slots+numSlots(bundle) > p.config.GlobalSlots {
-		return ErrBundleGasPriceLow
+	if p.slots+numSlots(bundle) > p.config.GlobalSlots {
```
welkin22, line 178: I think this section should be placed after line 186. After this bundle successfully drops some bundles, it is strange that it does not enter the bundle pool because it already exists.

redhdx: Make sense. Fixed.

redhdx:
> plz add UT for this feature.

Done. Plz review.

welkin22 (approved):
LGTM

owen-reorg (approved):

PR #140, opened 2024-08-15
perf: add DialOptions function for Dial

constwz:
### Description

add DialOptions function for Dial 

### Rationale

eoa-paymaster need to get opbnb block data continualy，so a long link needs to be created
eoa-paymaster is a service for gasless project


welkin22 (approved):
LGTM

andyzhang2023 (approved):
LGTM

PR #141, opened 2024-08-20
Fix: clear difflayer cache when truncate not triggered

krish-nr:
### Description

This PR aims to fix a potential memory leak issue in nodebuffer.

### Rationale

In #121 we introduced a potential memory leak risk. When the block height difference between stateid and truncate tail in the differlayer exceeds the limit size (90,000), we attempt to trigger a truncate. However, to prevent repeated deletions, we added a condition. If the condition is not met, we do not perform the truncate and return directly, waiting until the next trigger. Before returning, we forgot to clear the difflayer cache, which causes the difflayer to be unrecoverable, resulting in a memory leak. The diagram shows the monitoring of this part of the memory (pathdb_difflayer_hashcache_size).

The metric shows the memory usage of difflayer before and after enabling this fix.

image

### Example

N/A

### DIFF
--- a/trie/triedb/pathdb/disklayer.go
+++ b/trie/triedb/pathdb/disklayer.go
@@ -22,6 +22,8 @@ import (
 	"sync"
 
 	"github.com/VictoriaMetrics/fastcache"
+	"golang.org/x/crypto/sha3"
+
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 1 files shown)


will-2012 (approved):

owen-reorg (approved):

redhdx (approved):
LGTM

aweneagle (commented):
LGTM

PR #148, opened 2024-08-23
pevm: add a side slot to trigger next tx advance;

galaio:
### Description

pevm: add a side slot to trigger next tx advance;

### Changes

Notable changes:
* pevm: add a side slot to trigger next tx advance;
* ...


sunny2022da (approved):
LGTM

PR #150, opened 2024-08-26
txDAG transfer by gasless transaction (#28)

andyzhang2023:
* txDAG transfer

* encode/decode txDAG data with ABI

* set txDAG receiver to a special address

---------

Co-authored-by: andyzhang2023 <andyzhang2023@gmail.com>

merge conflict

fix conflict

### Description

### Rationale

### Example

### Changes

Notable changes:
* ...


miner/worker.go
```diff
@@ -893,10 +900,32 @@ func (w *worker) commitTransactions(env *environment, txs *transactionsByPriceAn
 	}
 	var coalescedLogs []*types.Log
 
+	//append the tx DAG transaction to the block
+	appendTxDAG := func() {
```
galaio, line 904: I think this method is better called in `generateWork`, It may be called multiple times in `commitTransactions`.

andyzhang2023: Indeed.

miner/worker.go
```diff
@@ -1167,6 +1246,8 @@ func (w *worker) fillTransactions(interrupt *atomic.Int32, env *environment) err
 			return err
 		}
 	}
+	// append a DAG tx at the end of the block
+	w.appendTxDAG(env)
```
galaio, line 1250: If there has MEV bundles, this tx may not the last one?

galaio (approved):

miner/worker.go
andyzhang2023: Yes, indeed.

PR #156, opened 2024-08-29
txdag: opt generation performance;

galaio:
### Description

mvstates: abandon RWKey, using origin key to record rwset
mvstates: support txdepMap to produce txdag;
mvststes: using struct in rwset;

### Changes

Notable changes:
* ...


andyzhang2023 (approved):
LGTM

PR #157, opened 2024-08-30
feat: cache lastTail in txindexer to avoid read from db

bnoieh:
### Description
This PR fix `engine_forkchoiceUpdated` api performance issue, which is introduced from upstream geth code.

### Rationale
1. `engine_forkchoiceUpdated` api blocks on [chainHeadFeed.Send](https://github.com/bnoieh/op-geth/blob/732b95e70d919dfa7ee26010fdb5561e09ee168b/core/blockchain.go#L2460), which waits for all the feed subscribers to receive.
2. [txIndexer](https://github.com/bnoieh/op-geth/blob/0085f637eb420dbc593b4d1bd06c66e77241a6f5/core/txindexer.go#L137) is one of subscribers, it may take about 800ms to receive, and the bad apple is `rawdb.ReadTxIndexTail`. You could [find more details here](https://github.com/ethereum/go-ethereum/issues/28907)
3. opBNB block time is 1s, and can not afford this high latency of engine api

### Example
n/a
### Changes
1. cache `lastTail` in `txindexer.go` to avoid read from db

### DIFF
--- a/core/rawdb/chain_iterator.go
+++ b/core/rawdb/chain_iterator.go
@@ -178,7 +178,7 @@ func iterateTransactions(db ethdb.Database, from uint64, to uint64, reverse bool
 //
 // There is a passed channel, the whole procedure will be interrupted if any
 // signal received.
-func indexTransactions(db ethdb.Database, from uint64, to uint64, interrupt chan struct{}, hook func(uint64) bool, report bool) {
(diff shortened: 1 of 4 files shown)


redhdx (approved):
LGTM

owen-reorg (approved):

redhdx (approved):

PR #169, opened 2024-09-10
use AddHash/ContainHash instead of Hasher interface

rickyyangz:
This change switches from using the `Hasher` interface to add/query the bloomfilter to implementing it as methods. This significantly reduces the allocations for Search and Rebloom.

### Description

### Rationale

### Example

### Changes

Notable changes:
* ...


redhdx (approved):

owen-reorg:
Close due to the ineffectiveness in the performance test.

PR #170, opened 2024-09-10
adjust bloomfilter parameters

rickyyangz:
### Description

### Rationale

### Example

### Changes

Notable changes:
* ...


redhdx:
Is there some data to support whether changes to these values ​​are reasonable?

owen-reorg:
adjusted in https://github.com/bnb-chain/op-geth/pull/230

PR #179, opened 2024-09-14
opt: tune bloom filter params of snapshot to fit higher tps

ghost:
### Description

tune bloom filter params of snapshot to fit higher tps
### Rationale

original bloom filter size is too small for opBNB 100m block gas limit, the error rate is 1 during load test.
we scale the size to fit opBNB throughput.
### Example

n/a
### Changes

Notable changes:
1. tune bloom filter params in `difflayer.go`


aweneagle (approved):
LGTM

PR #184, opened 2024-09-24
txdag: Add the dependency of read-before-write to the DAG

welkin22:
### Description

We need to ensure that write transactions on the data occur after read transactions. Therefore, if there is a read transaction before a write transaction, the read transaction will be added to the dependency list.


sunny2022da (approved):
lgtm

galaio (approved):

PR #186, opened 2024-09-25
fix: dump command

Eoous:
### Description

Currently `dump` command would close db after creating it immediately

### Rationale

`dump` command is not working


owen-reorg:
Thank you for your contribution. We'll leave it for now and merge the upstream fix https://github.com/ethereum-optimism/op-geth/commit/28d55218f7d793c184f4220a16a60e309caa70af to reduce further conflict.

owen-reorg:
If you do need to use the dump command now, feel free to use the this fixed version.

PR #188, opened 2024-09-26
txdag: optimize the size of readSet

welkin22:
### Description

When a read operation is accompanied by a write operation, there is no need to include it in the readSet because it will be filtered out later. We only need to record pure read operations.


sunny2022da (approved):
LGTM

0xbundler (approved):

PR #195, opened 2024-10-11
PEVM-fix: check in ValidatePlainTxDAG

sunny2022da:
### Description

### Rationale

### Example

### Changes

Notable changes:
* ...


welkin22 (approved):
LGTM

PR #199, opened 2024-10-17
Fix/pevm v0.5.0 invalid gas used

andyzhang2023:
### Description
bugfix: 'invalid gas used'
    1. add testcases for selfdestruct and create, which focus on balance
    2. add unit test for the false cache, which lead to 'invalid gas used'
    3. fix the false state of cache in uncommittedDB
    4. add a unit test case for GetCode()


sunny2022da (approved):
LGTM

welkin22 (approved):
LGTM

PR #200, opened 2024-10-17
rename pevm option

andyzhang2023:
do some args optimizations:
1. '--parallel2' is remove.
2. '--parallel'  will be used instead of '--parallel2'.
3. '--parallel-legacy' is added for the old version of pevm.
4. add log info when enabling parallel evm.


sunny2022da (approved):
LGTM

welkin22 (approved):
LGTM

PR #202, opened 2024-10-21
pevm: unordered merge mode flag

welkin22:
### Description

Add an unordered merge mode and include a flag to enable it.

### Rationale

In the confirm phase of PEVM, we fully trust the txDAG. The advantage of this is that we can skip conflict checks and immediately merge the execution results into the statedb after executing transactions at each txlevel. This will achieve better performance.

### Example

--parallel.unordered-merge

### Changes

Notable changes:
* Added the parallel.unordered-merge flag
* Add some metrics

### DIFF
--- a/core/pevm_processor.go
+++ b/core/pevm_processor.go
@@ -35,7 +35,7 @@ func newPEVMProcessor(config *params.ChainConfig, bc *BlockChain, engine consens
 		unorderedMerge: bc.vmConfig.EnableParallelUnorderedMerge,
 	}
 	log.Info("Parallel execution mode is enabled", "Parallel Num", ParallelNum(),
-		"CPUNum", runtime.NumCPU())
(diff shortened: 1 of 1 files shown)


sunny2022da (approved):
LGTM

core/state/pevm_statedb.go
```diff
@@ -486,9 +486,9 @@ func (pst *UncommittedDB) Merge(deleteEmptyObjects bool) error {
 		// so we don't need to merge anything.
 		return nil
 	}
-	if err := pst.conflictsToMaindb(); err != nil {
-		return err
-	}
+	//if err := pst.conflictsToMaindb(); err != nil {
+	//	return err
+	//}
```
sunny2022da, line 491: we can keep this uncommented, and make separate PR later for this when it is confirmed to be OK

PR #203, opened 2024-10-22
pevm: If TxDAG is nil, then use the serial processor to handle it.

welkin22:
### Description

If TxDAG is nil, then use the serial processor to handle it.

### Rationale

The current version of TxDAG reads from a file, so it may not be able to read after a certain block height, and a serial processor should be used to execute it.


sunny2022da (approved):
LGTM!
maybe in future we can have a separate flag such as "parallel.trustDAG", so that parallel merge and other features could reuse it . 
But let's keep the minimal change at present.

PR #206, opened 2024-10-25
bugfix: tstorage revert and selfdestruct6780 revert

andyzhang2023:
### Description

The case of `go test -run TestExecutionSpecBlocktests` failed when run in `tests` dir. It said: `invalid gas used` and `invalid mercle root`.

### Rationale
There are two bugs which caused these:
1. The pre-value is incorrect when adding journal for TransientStorage. It should be the previous one, but not the current one.
2. The op SelfDestruct6780 forgot to add its journal for revert. It should call SelfDestruct() to do it.

### DIFF
--- a/core/parallel_state_scheduler.go
+++ b/core/parallel_state_scheduler.go
@@ -12,19 +12,22 @@ import (
 )
 
 var runner chan func()
+var runnerOnce sync.Once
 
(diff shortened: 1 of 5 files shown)


sunny2022da (approved):
LGTM

welkin22 (approved):
LGTM

PR #207, opened 2024-10-25
increase the conflict counter only if the state of uncommitted db con…

andyzhang2023:
### Description

increase the conflict counter only if the state of uncommitted db conflicts to maindb


welkin22 (approved):
LGTM

sunny2022da (approved):
LGTM

PR #210, opened 2024-10-31
Fix/pevm oom

andyzhang2023:
### Description

There is an oom issue in transaction pool brought by reheap optimization:
1. The demoted transactions would not be removed unless the pool is full, which means those staled transactions will stay in the priced list forever until the pool overflows.
2. memory grows too fast if the pool copy all transactions 5 times per second, so we reduce it to 1 time per second.


sunny2022da (commented):
LGTM，suggested to submit to develop branch and PEVM can cherry-pick.

PR #212, opened 2024-10-31
pevm: refactor txdag reader

welkin22:
### Description

The process of the txDAG reader reading files should not occur during block processing. It should be handled separately by a goroutine to achieve better performance and lower memory usage.

### Rationale

Refactored the implementation of txDAGReader to improve performance and reduce memory usage.

### Example

none

### Changes

Notable changes:
-  Refactored txDAGReader and moved the related code to a separate txdag_reader.go file
-  Fixed the related unit tests


sunny2022da (approved):
LGTM

PR #214, opened 2024-11-01
pevm: fix the blockNumber used by txDAGReader during initialization

welkin22:
### Description

Use bc.CurrentBlock() instead of bc.CurrentHeader() when initializing txDAGReader.

### Rationale

When a block rollback occurs, using bc.CurrentBlock() is the way to get the correct starting block height.

### Example

none


sunny2022da (approved):
LGTM

PR #218, opened 2024-11-08
PEVM-fix: Disable prefetch when PEVM enabled

sunny2022da:
There is no necessary to do prefetch for PEVM.

### Description

### Rationale

### Example

### Changes

Notable changes:
* ...


andyzhang2023 (approved):
LGTM

welkin22 (approved):
LGTM

PR #127, opened 2024-07-10
feat: support multi database feature for op

jingjunLi:
### Description

The principle support of PR: [BEP-365: Support multi-database based on data pattern](https://forum.bnbchain.org/t/bep-365-support-multi-database-based-on-data-pattern/2655)

### Rationale

tell us why we need these changes...

### Example

add an example CLI or API response...

### Changes

Notable changes:
* add each change in a bullet point here
* ...

PR #220, opened 2024-11-12
fix: fix StateScheme overwrite bug

jingjunLi:
### Description

In PR #127, while resolving merge conflicts, the return value of ParseStateScheme was unintentionally altered. This change caused the config.StateScheme variable to be consistently set to nil when passed to pbss, leading to issues in downstream operations.

### Rationale

Restored the correct return value in ParseStateScheme to ensure that config.StateScheme is properly assigned and passed to pbss, restoring intended functionality.

### Example

### Changes

Notable changes:
* ...


krish-nr (approved):
LGTM

PR #224, opened 2024-11-15
fix: recover node buffer list trie nodes for graceful kill

sysvm:
### Description

This pr aims to recover node buffer list data for graceful kill. Therefore, both graceful kill and force kill can recover data by freezerdb.

### Rationale

When using graceful kill to stop geth, the data in nodebufferlist is stored in db or journalFile. When nodebufferlist recovers data by loading all data into nodebufferlist base buffer, if the data is too large, it will exceed the pebble batch size limit and cause panic. By this pr, both graceful kill and force kill can use the same functions to recover data.

### Example

N/A

### Changes

Notable changes:
* N/A


krish-nr (dismissed):
LGTM

joey0612 (approved):

will-2012 (approved):

PR #225, opened 2024-11-19
feat: PEVM fallback to serial process

sunny2022da:
fallback to serial processor when pevm execution result is not correct

### Description

### Rationale

### Example

### Changes

Notable changes:
* ...

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1953,8 +1953,15 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool) (int, error)
 				blockProcessedInParallel = true
 				if err != nil {
 					// parallel processing fail , fallback to serial with new statDB.
-					log.Warn("ParallelEVM fallback!!!", "error", err.Error())
+					log.Warn("ParallelEVM fallback to serial process", "error", err.Error())
+					execErr := err
 					statedb, err = bc.reGenerateStateForFallBack(parent.Root, block.Root(), statedb)
+					if err != nil {
+						// Can not get new statedb for serial run, report the process error.
+						bc.reportBlock(block, receipts, execErr)
+						followupInterrupt.Store(true)
+						return it.index, err
+					}
 					statedb.StartPrefetcher("chain")
 					activeState = statedb
 					receipts, logs, usedGas, err = bc.serialProcessor.Process(block, statedb, bc.vmConfig)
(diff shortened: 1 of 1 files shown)


core/blockchain.go
```diff
@@ -1940,8 +1947,19 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool) (int, error)
...
 			if bc.vmConfig.TxDAG == nil && bc.vmConfig.EnableParallelUnorderedMerge {
 				receipts, logs, usedGas, err = bc.serialProcessor.Process(block, statedb, bc.vmConfig)
+				blockProcessedInParallel = false
 			} else {
 				receipts, logs, usedGas, err = bc.processor.Process(block, statedb, bc.vmConfig)
+				blockProcessedInParallel = true
+				if err != nil {
+					// parallel processing fail , fallback to serial with new statDB.
+					log.Warn("ParallelEVM fallback!!!", "error", err.Error())
+					statedb, err = bc.reGenerateStateForFallBack(parent.Root, block.Root(), statedb)
```
welkin22, line 1957: The error should be handled here because if err is not nil, then statedb will be nil. In this case, the next line `statedb.StartPrefetcher` could potentially cause a panic.

core/blockchain.go
```diff
@@ -1953,9 +1971,30 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool) (int, error)
...
-			return it.index, err
+			if blockProcessedInParallel {
+				// invalid parallel execution, try serial
+				log.Warn("ParallelEVM fallback after ValidateState!!!", "error", err.Error())
+				parent := it.previous()
+				if parent == nil {
+					parent = bc.GetHeader(block.ParentHash(), block.NumberU64()-1)
+				}
+
+				statedb, err = bc.reGenerateStateForFallBack(parent.Root, block.Root(), statedb)
```
welkin22, line 1982: Error should also be handled here.

core/blockchain.go
DavidZangNR: Done

core/blockchain.go
DavidZangNR: Done

andyzhang2023 (approved):
LGTM

welkin22 (approved):
LGTM

PR #230, opened 2024-11-29
perf: tune snapshot bloom for higher tps

bnoieh:
### Description

tune bloom filter params of snapshot to fit higher tps

### Rationale

original bloom filter size is too small for opBNB 100m block gas limit, the error rate is 1 during load test.
we adjust it to fit opBNB throughput.
### Example

n/a
### Changes
1. adjust the bloom size calculation to include all diffLayers and aggregatorLayer
2. scale up the size for bigger block gas limit of opBNB
3. scale down the error target rate 0.02 -> 0.01


andyzhang2023 (approved):
LGTM

krish-nr (approved):
LGTM

PR #238, opened 2024-12-12
pevm: support parallel merge mode

welkin22:
### Description

Currently, in our PEVM code, only the execution phase is parallelized. After the parallel execution ends, the results need to be merged serially into the StateDB. I have modified the merging process to be executed in parallel as well, which further improves the performance of the PEVM and reduces the time consumption.

### Rationale

By providing a new ParallelStateDB to replace the original StateDB, it supports concurrent reading and writing of data.
Provide ParallelGasPool instead of the original GasPool to support gas calculation in a concurrent environment.
Other similar changes.

### Example
Enable parallel merge mode by adding startup parameters:
```
--parallel.parallel-merge
```
Note that this startup parameter conflicts with parallel.unordered-merge, and when both are set, parallel.unordered-merge will be invalid.

### Changes

Notable changes:
* ParallelStateDB
* ParallelGasPool
* The `state.StateDBer` interface replaces direct references to `StateDB`.
* I supplemented a large number of txDAG files for the `TestBlockchainWithTxDAG` unit test. I found that the previous unit test had issues, as the code did not enter the branch that uses the DAG files.


core/blockchain.go
```diff
@@ -1974,22 +1978,28 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool) (int, error)
 			if parent == nil {
 				parent = bc.GetHeader(block.ParentHash(), block.NumberU64()-1)
 			}
-			statedb, err = state.New(parent.Root, bc.stateCache, bc.snaps)
+			// Enable prefetching to pull in trie node paths while processing transactions
```
sunny2022da, line 1981: This comments should be at line 1996?

sunny2022da (commented):
Generally LGTM, just a minor change needed.

core/blockchain.go
welkin22: yes, already fixed

sunny2022da (approved):
LGTM

andyzhang2023 (approved):
LGTM

PR #240, opened 2024-12-18
feat(op-geth): support droppingTxHashes when sendBundle

redhdx:
### Description

Support droppingTxHashes when sendBundle.

### Rationale

The atomicity of bundle transactions can lead to a malicious attack. The attacker can send a failed transaction, such as a transaction that has already been on the chain, which will make the entire bundle invalid. Therefore, the new field can remove such transactions.

### Example

N/A

### Changes

Notable changes:
* Add new field in Bundle struct.

### DIFF
--- a/core/types/transaction.go
+++ b/core/types/transaction.go
@@ -632,6 +632,12 @@ func (s Transactions) EncodeIndex(i int, w *bytes.Buffer) {
 	}
 }
 
+func (s Transactions) Remove(idx int) Transactions {
+	copy(s[idx:], s[idx+1:])
+	s[len(s)-1] = nil
+	return s[:len(s)-1]
+}
+
 // TxDifference returns a new set which is the difference between a and b.
 func TxDifference(a, b Transactions) Transactions {
 	keep := make(Transactions, 0, len(a))
--- a/miner/worker_builder.go
+++ b/miner/worker_builder.go
@@ -366,13 +366,25 @@ func (w *worker) simulateBundle(
 		bundleGasFees = new(big.Int)
 	)
 
+	currentState := state.Copy()
+
 	for i, tx := range bundle.Txs {
 		state.SetTxContext(tx.Hash(), i+currentTxCount)
+
+		prevState := currentState.Copy()
+		prevGasPool := new(core.GasPool).AddGas(gasPool.Gas())
 
 		receipt, err := core.ApplyTransaction(w.chainConfig, w.chain, &w.coinbase, gasPool, state, env.header, tx,
 			&tempGasUsed, *w.chain.GetVMConfig())
 		if err != nil {
 			log.Warn("fail to simulate bundle", "hash", bundle.Hash().String(), "err", err)
+			if containsHash(bundle.DroppingTxHashes, tx.Hash()) {
+				log.Warn("drop tx in bundle", "hash", tx.Hash().String())
+				state = prevState
+				gasPool = prevGasPool
+				bundle.Txs = bundle.Txs.Remove(i)
+				continue
+			}
 
 			if prune {
 				if errors.Is(err, core.ErrGasLimitReached) && !pruneGasExceed {
@@ -387,6 +399,14 @@ func (w *worker) simulateBundle(
 		}
 
 		if receipt.Status == types.ReceiptStatusFailed && !containsHash(bundle.RevertingTxHashes, receipt.TxHash) {
+			// for unRevertible tx but itself can be dropped, we drop it and revert the state and gas pool
+			if containsHash(bundle.DroppingTxHashes, receipt.TxHash) {
+				log.Warn("drop tx in bundle", "hash", receipt.TxHash.String())
+				state = prevState
+				gasPool = prevGasPool
+				bundle.Txs = bundle.Txs.Remove(i)
+				continue
+			}
 			err = errNonRevertingTxInBundleFailed
 			log.Warn("fail to simulate bundle", "hash", bundle.Hash().String(), "err", err)
 
@@ -412,6 +432,14 @@ func (w *worker) simulateBundle(
 			bundleGasFees.Add(bundleGasFees, txGasFees)
 		}
 	}
+
+	// prune bundle when all txs are dropped
+	if len(bundle.Txs) == 0 {
+		log.Warn("prune bundle", "hash", bundle.Hash().String(), "err", "empty bundle")
+		w.eth.TxPool().PruneBundle(bundle.Hash())
+		return nil, errors.New("empty bundle")
+	}
+
 	// if all txs in the bundle are from txpool, we accept the bundle without checking gas price
 	bundleGasPrice := big.NewInt(0)
 	if bundleGasUsed != 0 {
--- a/core/types/bundle.go
+++ b/core/types/bundle.go
@@ -23,6 +23,7 @@ type SendBundleArgs struct {
 	MinTimestamp      *uint64         `json:"minTimestamp"`
 	MaxTimestamp      *uint64         `json:"maxTimestamp"`
 	RevertingTxHashes []common.Hash   `json:"revertingTxHashes"`
+	DroppingTxHashes  []common.Hash   `json:"droppingTxHashes"`
 }
 
 type Bundle struct {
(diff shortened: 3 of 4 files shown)


irrun (approved):

miner/worker_builder.go
```diff
@@ -366,13 +366,25 @@ func (w *worker) simulateBundle(
 		bundleGasFees = new(big.Int)
 	)
 
+	currentState := state.Copy()
+
 	for i, tx := range bundle.Txs {
 		state.SetTxContext(tx.Hash(), i+currentTxCount)
 
+		prevState := currentState.Copy()
```
welkin22, line 374: state.Copy is a very performance-intensive action. If you want it to revert to the previous state when a transaction fails, you should use snapshot-related methods. Otherwise, performance will significantly degrade.

miner/worker_builder.go
```diff
@@ -366,13 +366,25 @@ func (w *worker) simulateBundle(
...
+
 		receipt, err := core.ApplyTransaction(w.chainConfig, w.chain, &w.coinbase, gasPool, state, env.header, tx,
 			&tempGasUsed, *w.chain.GetVMConfig())
 		if err != nil {
 			log.Warn("fail to simulate bundle", "hash", bundle.Hash().String(), "err", err)
+			if containsHash(bundle.DroppingTxHashes, tx.Hash()) {
+				log.Warn("drop tx in bundle", "hash", tx.Hash().String())
+				state = prevState
+				gasPool = prevGasPool
+				bundle.Txs = bundle.Txs.Remove(i)
```
owen-reorg, line 385: Remove with index in the iteration may cause issue.
We should mark and rebuild the tx slice later to avoid it.
Plz check the PoC below.

https://go.dev/play/p/NlIeCifQcfY

miner/worker_builder.go
redhdx: Fixed. Please review again.

miner/worker_builder.go
redhdx: Fixed. Please review again. We can't simply continue without modifying the txs, we must remove the invalid tx or commit bundle when building block will fail.

core/types/transaction.go
```diff
@@ -632,6 +632,12 @@ func (s Transactions) EncodeIndex(i int, w *bytes.Buffer) {
 	}
 }
 
+func (s Transactions) Remove(idx int) Transactions {
+	copy(s[idx:], s[idx+1:])
```
owen-reorg, line 636: It works, but not memory efficient. I would suggest a method as below.

```
// removeElements removes elements from a slice based on a given condition
// Returns a new slice containing only the elements that don't match the condition
// This method modifies the original slice and is memory efficient
func removeElements(slice []int, condition func(int) bool) []int {
    j := 0
    for i := 0; i < len(slice); i++ {
        if !condition(slice[i]) {
            // Only keep elements that don't match the removal condition
            // If i and j are different, move the element to position j
            if i != j {
                slice[j] = slice[i]
            }
            j++ // Increment j only when we keep an element
        }
    }
    // Return the slice up to position j, which contains all kept elements
    return slice[:j]
}

// Usage example
func main() {
    numbers := []int{1, 2, 3, 4, 5}
    // Remove even numbers from the slice
    result := removeElements(numbers, func(n int) bool {
        return n%2 == 0
    })
    fmt.Println(result) // Output: [1 3 5]
}

// Key benefits of this approach:
// 1. Single pass through the slice
// 2. No additional memory allocation
// 3. In-place modification
// 4. Time complexity: O(n)
// 5. Space complexity: O(1)
```

core/txpool/bundlepool/bundlepool.go
```diff
@@ -370,16 +374,27 @@ func (p *BundlePool) reset(newHead *types.Header) {
...
+						p.slots -= numSlots(p.bundles[hash])
+						delete(p.bundles, hash)
+						break
+					}
+				}
+			}
+		}(hash, bundle)
 	}
+	wg.Wait()
 	bundleGauge.Update(int64(len(p.bundles)))
```
owen-reorg, line 398: do we need these 2 lines if we early return when len(p.bundles) == 0?

core/txpool/bundlepool/bundlepool.go
```diff
@@ -370,16 +374,27 @@ func (p *BundlePool) reset(newHead *types.Header) {
...
-		} else if txSet.Contains(bundle.Txs[0].Hash()) {
-			p.slots -= numSlots(p.bundles[hash])
-			delete(p.bundles, hash)
-		}
+		wg.Add(1)
+		go func(hash common.Hash, bundle *types.Bundle) {
+			defer wg.Done()
+			if (bundle.MaxTimestamp != 0 && newHead.Time > bundle.MaxTimestamp) ||
+				(bundle.MaxBlockNumber != 0 && newHead.Number.Cmp(new(big.Int).SetUint64(bundle.MaxBlockNumber)) > 0) {
+				p.slots -= numSlots(p.bundles[hash])
```
owen-reorg, line 384: Is it thread safe to do `p.slots -=` and `delete(p.bundles, hash)`?

core/txpool/bundlepool/bundlepool.go
```diff
@@ -370,16 +374,27 @@ func (p *BundlePool) reset(newHead *types.Header) {
...
+				delete(p.bundles, hash)
+			} else {
+				for _, tx := range bundle.Txs {
+					if txSet.Contains(tx.Hash()) && !containsHash(bundle.DroppingTxHashes, tx.Hash()) {
```
owen-reorg, line 388: I recommend maintain unDroppableTxHashesSet as a cache, we can use it here and in the check of simulation, avoid iterate the slice again and again.

miner/worker_builder.go
```diff
@@ -366,13 +366,24 @@ func (w *worker) simulateBundle(
 		bundleGasFees = new(big.Int)
 	)
 
-	for i, tx := range bundle.Txs {
-		state.SetTxContext(tx.Hash(), i+currentTxCount)
+	succeedTxs := types.Transactions{}
+	succeedTxCount := 0
```
owen-reorg, line 370: It seems useless.

miner/worker_builder.go
```diff
@@ -387,6 +398,11 @@ func (w *worker) simulateBundle(
 		}
 
 		if receipt.Status == types.ReceiptStatusFailed && !containsHash(bundle.RevertingTxHashes, receipt.TxHash) {
+			// for unRevertible tx but itself can be dropped, we drop it and revert the state and gas pool
+			if containsHash(bundle.DroppingTxHashes, receipt.TxHash) {
+				log.Warn("drop tx in bundle", "hash", receipt.TxHash.String())
```
owen-reorg, line 403: We should also do this here?
```
state.RevertToSnapshot(snap)
gasPool.SetGas(gp)
```

core/txpool/bundlepool/bundlepool.go
owen-reorg: The `reset` function itself is thread safe since we have 
```
	p.mu.Lock()
	defer p.mu.Unlock()
```
in the beginning of the function.

However, if you use go func here, it's not.

miner/worker_builder.go
redhdx: We can't do this here beacause in the evm execution, when status was failed, it was already reverted. If revert here, it will panic.

core/txpool/bundlepool/bundlepool.go
redhdx: Added.

core/txpool/bundlepool/bundlepool.go
redhdx: Removed go func, just use single thread.

miner/worker_builder.go
irrun: The actual reason for no need to revert state here is that, when no err returned by ApplyTransaction, state.clearJournalAndRefund() must had been called by  https://github.com/bnb-chain/op-geth/blob/develop/core/state_processor.go#L167 or https://github.com/bnb-chain/op-geth/blob/develop/core/state/statedb.go#L1008 to avoid reverting across transactions, after that, `state.validRevisions` would be clear, so if here do revert, panic must happen for the first check is true at https://github.com/bnb-chain/op-geth/blob/develop/core/state/statedb.go#L918, so we directly remove the tx from bundle.

PR #242, opened 2024-12-25
 Revert "optimization: txpool pending cache improvement (#177)"

andyzhang2023:
### Description

The improvement of #177 saves about 50+ms of calling txpool.Pending() for the miner, when the pending transactions accumulate up to 100k+.  
**The improvement of performance depends on the number of pending list.**
However, after we had resolved other bottlenecks, the pending transactions will usually be under 100k+, which means it can't save more than 50+ms, and besides, it changes a lot of code to get the improvement. 
So we decided to revert it and make it a draft, to re-enable it in the future when needed.


andyzhang2023:
This PR is relied on and included by  #244 , so we drop it.

PR #243, opened 2024-12-25
Txpool optimization: broadcast transaction body directly to peer

andyzhang2023:
### Description
There are two ways of broadcasting transactions to peers:  1. broadcasting bodies; 2. Announcing hashes.
The former is more efficient, by which we transfer all transactions when supporting a high TPS throughput. It can be enabled by flag **--p2p.directly** now.


andyzhang2023:
> i suggest change the flag from bool to []broadCastPeers, b/c the peers need to broadcast directly is not exactly equal to static peers in many cases.

It's a good idea. Replaced.

eth/handler.go
```diff
@@ -620,6 +630,29 @@ func (h *handler) BroadcastBlock(block *types.Block, propagate bool) {
 	}
 }
 
+func (h *handler) peersForBroadcasting(numDirect int, peers []*ethPeer) (direct []*ethPeer, announce []*ethPeer) {
+	// Split the peers into direct-peers and announce-peers
+	// we send the tx directly to direct-peers; all static nodes are direct-peers
```
owen-reorg, line 635: `all static nodes are direct-peers` is not true anymore if we have direct nodes parameter.

andyzhang2023: comment fixed.

owen-reorg (dismissed):

PR #245, opened 2024-12-25
Txpool optimization: disable cache if --mine is not enabled

andyzhang2023:
### Description
This is an improvement based on [#85](https://github.com/bnb-chain/op-geth/pull/85).
The pending cache optimization [#85](https://github.com/bnb-chain/op-geth/pull/85) only works on the miner. It doesn't work on the followers, and it costs memory. So we disable it if the node is not a miner.


core/txpool/legacypool/legacypool.go
```diff
@@ -645,35 +635,57 @@ func (pool *LegacyPool) ContentFrom(addr common.Address) ([]*types.Transaction,
 // The transactions can also be pre-filtered by the dynamic fee components to
 // reduce allocations and load on downstream subsystems.
 func (pool *LegacyPool) Pending(filter txpool.PendingFilter) map[common.Address][]*txpool.LazyTransaction {
-	empty := txpool.PendingFilter{}
-	if filter == empty {
-		// return all pending transactions, no filtering
-		return pool.pendingCache.dump(false)
-	}
+	// TODO need to confirm
```
owen-reorg, line 638: plz resolve this comment

andyzhang2023: It's a legacy comment of "lazy transaction optimization", which is reverted now. So we don't need it any more.

owen-reorg (approved):

PR #246, opened 2024-12-26
Txpool opt async priced

andyzhang2023:
### Description
An "async-reheap" priced list is introduced into txpool to get better performance. The re-heap of the whole transaction pool usually costs hundreds of ms, which blocks more transactions coming into the pool through the Add() function. So we develop a new type of priced list to resolve the bottleneck. 
The "async-reheap" finds the "lowest-priced" transactions async, and stores it into a cache, then it  does the "underpriced" validation quickly by the stored cache.

### Example

add flag ```--txpool.asyncpriced``` to enable this feature.

### DIFF
--- a/core/txpool/legacypool/async_priced_list.go
+++ b/core/txpool/legacypool/async_priced_list.go
@@ -149,6 +149,8 @@ func (a *asyncPricedList) Underpriced(tx *types.Transaction) bool {
 		// be careful that fl might be nil
 		floatingLowest = fl.(*types.Transaction)
 	}
+	a.mu.Lock()
+	defer a.mu.Unlock()
 	return (urgentLowest == nil || a.priced.urgent.cmp(urgentLowest, tx) >= 0) &&
 		(floatingLowest == nil || a.priced.floating.cmp(floatingLowest, tx) >= 0) &&
 		(floatingLowest != nil || urgentLowest != nil)
--- a/core/txpool/legacypool/async_priced_list.go
+++ b/core/txpool/legacypool/async_priced_list.go
@@ -1,0 +1,174 @@
+package legacypool
+
+import (
+	"math/big"
+	"sync"
+	"sync/atomic"
+
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+var _ pricedListInterface = &asyncPricedList{}
+
+type addEvent struct {
+	tx    *types.Transaction
+	local bool
+}
+
+type discardEvent struct {
+	slots int
+	force bool
+	done  chan *discardResult
+}
+type discardResult struct {
+	discardTxs types.Transactions
+	succ       bool
+}
+
+type asyncPricedList struct {
+	priced         *pricedList
+	floatingLowest atomic.Value
+	urgentLowest   atomic.Value
+	mu             sync.Mutex
+
+	// events
+	quit       chan struct{}
+	reheap     chan struct{}
+	add        chan *addEvent
+	remove     chan int
+	discard    chan *discardEvent
+	setBaseFee chan *big.Int
+}
+
+func newAsyncPricedList(all *lookup) *asyncPricedList {
+	a := &asyncPricedList{
+		priced:     newPricedList(all),
+		quit:       make(chan struct{}),
+		reheap:     make(chan struct{}),
+		add:        make(chan *addEvent),
+		remove:     make(chan int),
+		discard:    make(chan *discardEvent),
+		setBaseFee: make(chan *big.Int),
+	}
+	go a.run()
+	return a
+}
+
+// run is a loop that handles async operations:
+//   - reheap: reheap the whole priced list, to get the lowest gas price
+//   - put: add a transaction to the priced list
+//   - remove: remove transactions from the priced list
+//   - discard: remove transactions to make room for new ones
+func (a *asyncPricedList) run() {
+	var reheap bool
+	var newOnes []*types.Transaction
+	var toRemove int = 0
+	// current loop state
+	var currentDone chan struct{} = nil
+	var baseFee *big.Int = nil
+	for {
+		if currentDone == nil {
+			currentDone = make(chan struct{})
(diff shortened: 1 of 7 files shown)


core/txpool/legacypool/legacypool.go
```diff
@@ -1462,11 +1459,13 @@ func (pool *LegacyPool) runReorg(done chan struct{}, reset *txpoolResetRequest,
...
 		pool.demoteUnexecutables(demoteAddrs)
 		demoteTimer.UpdateSince(t0)
-		var pendingBaseFee = pool.priced.urgent.baseFee
+		var pendingBaseFee = pool.priced.GetBaseFee()
 		if reset.newHead != nil {
 			if pool.chainconfig.IsLondon(new(big.Int).Add(reset.newHead.Number, big.NewInt(1))) {
 				pendingBaseFee = eip1559.CalcBaseFee(pool.chainconfig, reset.newHead, reset.newHead.Time+1)
 				pool.priced.SetBaseFee(pendingBaseFee)
+			} else {
+				pool.priced.Reheap()
```
owen-reorg, line 1468: If we only do reheap when runReorg instead of add, will the node OOM when there are tons of txns adding into txpool in a block interval?

core/txpool/legacypool/async_priced_list.go
```diff
@@ -0,0 +1,191 @@
...
+	"sync/atomic"
+
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+var _ pricedListInterface = &asyncPricedList{}
+
+type addEvent struct {
+	tx    *types.Transaction
+	local bool
```
owen-reorg, line 15: It seems `local` is not used, why do we bother put it here?

core/txpool/legacypool/async_priced_list.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+}
+
+func (a *asyncPricedList) handle(reheap bool, newOnes []*types.Transaction, toRemove int, baseFee *big.Int, finished chan struct{}) {
+	defer close(finished)
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	// add new transactions to the priced list
+	for _, tx := range newOnes {
+		a.priced.Put(tx, false)
```
owen-reorg, line 104: we just make all txns `local` as false and ignored the one passed with `AddEvent`. Is it expected?

core/txpool/legacypool/legacypool.go
aweneagle: It will be limited by the TPS of txpool

core/txpool/legacypool/async_priced_list.go
aweneagle: "asyncPricedList" is a wrap of the original one "pricedList", so when we call Add(tx, local) on asyncPricedList, we just call a "asynced" Add(tx, local) pricedList. And the `local' here is for passing the parameter to pricedList.

aweneagle: "asyncPricedList" is a wrap of the original one "pricedList", so when we call Add(tx, local) on asyncPricedList, we just call a "asynced" Add(tx, local) of pricedList. And the `local' here is for passing the parameter to pricedList.

core/txpool/legacypool/async_priced_list.go
aweneagle: Yes, It's ok because the pricedList will also ignore those transactions that local=true.

core/txpool/legacypool/legacypool.go
andyzhang2023: It will be limited by the TPS of txpool

core/txpool/legacypool/async_priced_list.go
andyzhang2023: "asyncPricedList" is a wrap of the original one "pricedList", so when we call Add(tx, local) on asyncPricedList, we just call a "asynced" Add(tx, local) of pricedList. And the `local' here is for passing the parameter to pricedList.

core/txpool/legacypool/async_priced_list.go
```diff
@@ -0,0 +1,191 @@
...
+
+func (a *asyncPricedList) Put(tx *types.Transaction, local bool) {
+	a.add <- &addEvent{tx, local}
+}
+
+func (a *asyncPricedList) Removed(count int) {
+	a.remove <- count
+}
+
+func (a *asyncPricedList) Underpriced(tx *types.Transaction) bool {
```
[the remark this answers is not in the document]
andyzhang2023: Yes, It's ok because the pricedList will also ignore those transactions that local=true.

core/txpool/legacypool/async_priced_list.go
andyzhang2023: Yes, It's ok because the pricedList will also ignore those transactions that local=true.

core/txpool/legacypool/list.go
```diff
@@ -549,7 +564,6 @@ func (h *priceHeap) Pop() interface{} {
 // better candidates for inclusion while in other cases (at the top of the baseFee peak)
 // the floating heap is better. When baseFee is decreasing they behave similarly.
 type pricedList struct {
-	currHead *types.Header // Current block header for effective tip calculation
```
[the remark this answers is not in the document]
andyzhang2023: It's reverted to the original pricedList. So we don't need the `currHead` anymore.

core/txpool/legacypool/async_priced_list.go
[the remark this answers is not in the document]
andyzhang2023: The lowest-priced transaction is already held in atomic.Value; But yes, it should be better to hold a mu to do the pricedHeap.cmp.   Fixed.

andyzhang2023:
> I see no log for the new feature, better add proper debug/info logs

Added.

owen-reorg (approved):

PR #251, opened 2025-01-10
pevm: fallback to sequencial processor when the TxDAG is too deep

andyzhang2023:
### Description

A new option is added to improve the performance when the TxDAG contains a deep dependency chain. The performance might become poor when many transactions depend on each other in a block. So we add a threshold option under which the parallel EVM will be enabled, otherwise, the block will be processed by the original EVM in sequential.

### Example

Add flag ```--parallel.txdag-max-depth-ratio``` to enable this feature.  Default to be not enabled.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1138,8 +1138,8 @@ Please note that --` + MetricsHTTPFlag.Name + ` must be set to start the server.
 
 	ParallelTxDATMaxDepthRatioFlag = &cli.Float64Flag{
 		Name:     "parallel.txdag-max-depth-ratio",
-		Usage:    "A ratio to decide whether or not to execute transactions in parallel, it will fallback to sequencial processor if the depth is larger than this value (default = 0.9)",
-		Value:    0.9,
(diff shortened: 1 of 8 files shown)


cmd/utils/flags.go
```diff
@@ -1136,6 +1136,13 @@ Please note that --` + MetricsHTTPFlag.Name + ` must be set to start the server.
 		Category: flags.VMCategory,
 	}
 
+	ParallelTxDATMaxDepthRatioFlag = &cli.Float64Flag{
+		Name:     "parallel.txdag-max-depth-ratio",
+		Usage:    "A ratio to decide whether or not to execute transactions in parallel, it will fallback to sequencial processor if the depth is larger than this value (default = 1.0)",
+		Value:    1.0,
```
sunny2022da, line 1142: may be we can set the default value to 0.9? it seems 1.0 means there is almost 1 tx in every level. so PEVM is expected to degrade the performance in this case.

sunny2022da (commented):
Generally LGTM. just minor suggestion of the default threshold.

cmd/utils/flags.go
andyzhang2023: Yes, changed to 1.0.

welkin22 (approved):
LGTM

PR #252, opened 2025-01-11
fix: txpool concurrent map issue

andyzhang2023:
### Description

This PR is for fixing the issue of concurrent map reading and writing on the map of txpool.queue.
It was introduced by PR #247 , in which the numbers of pending and queued transactions were added to logs. 
The txpool.queue and txpool.pending should always be protected by the mu lock of txpool, So we use Stats() instead of stats();
Besides, there is no need to iterator the pending and queue lists to sum the total numbers. The variables pendingCounter and queueCounter records them.


krish-nr (approved):
LGTM

PR #256, opened 2025-01-16
Fix pevm 20per performance issue

andyzhang2023:
### Description
Some improvements are added in the PR:
1. fix the default value of flag ```--parallel.txdag-max-depth-ratio```
2. change the metric of "parallel txs depth" from counter to gauge.
3. change the maindb of UncommittedDB from ParallelStateDB to StateDB, which caused a 20%+ decrease of performance.


sunny2022da (approved):
LGTM

PR #258, opened 2025-01-20
fix(op-geth): fix bundle memory leak

redhdx:
### Description

Fix bundle memory leak.

### Rationale

The bundle pool has a memory leak problem, and the bundle heap has no reconstruction logic, which causes the bundle transactions to always fail to GC.

### Example

N/A.

### Changes

Notable changes:
* Add bundle heap reorg when reset bundle pool.

### DIFF
--- a/core/txpool/bundlepool/bundlepool.go
+++ b/core/txpool/bundlepool/bundlepool.go
@@ -361,6 +361,12 @@ func (p *BundlePool) reset(newHead *types.Header) {
 	p.mu.Lock()
 	defer p.mu.Unlock()
 
+	if len(p.bundles) == 0 {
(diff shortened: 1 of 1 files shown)


core/txpool/bundlepool/bundlepool.go
```diff
@@ -379,6 +386,9 @@ func (p *BundlePool) reset(newHead *types.Header) {
 			p.slots -= numSlots(p.bundles[hash])
 			delete(p.bundles, hash)
 		}
+		if p.bundles[hash] != nil {
+			p.bundleHeap.Push(bundle)
```
owen-reorg, line 390: Shall we use heap.push here?

Boomtokn (approved):
LGTM

PR #267, opened 2025-02-26
fix: computing unexpected block hash

Eoous:
### Description

add `RequestsHash` support then `Block.Hash()` returns correct hash

### Changes

Notable changes:

regenerated rlp for `Header`


will-2012:
Currently, op-geth does not support EIP-7685 as a whole. Why introduce this change?

Boomtokn (approved):
LGTM

PR #269, opened 2025-03-12
feat: change block internal to 500ms

joey0612:
### Background

Adapt to [BEP-524](https://github.com/buddh0/BEPs/blob/improve-block-producer-selection/BEPs/BEP-524.md), opBNB block interval will be shortened to 500ms. 

If the block time of opBNB is greater than the block time of BSC, over time, opBNB will never be able to sync to the latest cross-chain messages, and opBNB will also stop producing blocks. In [BEP-524](https://github.com/buddh0/BEPs/blob/improve-block-producer-selection/BEPs/BEP-524.md), the block time of BSC will be shortened to 750 ms, and before that, opBNB will shorten its block interval to 500 ms.

### Design

Considering compatibility, op-geth references the bytes from BSC to store the millisecond part in the first two parts of `MixDigest` in the block header; op-node will handle the setting of the millisecond part, while op-geth only needs to read it correctly. So, the timestamp in the block still represents a second-level timestamp.

In op-geth, there are two scenarios for accessing timestamps:
1. Based on the block timestamp, determine the fork.
2. In the miner and EngineAPI, it is necessary to convert to milliseconds to determine whether the relationship between the previous and current blocks is correct.

The former still uses a second-level timestamp, while the latter must have a method to obtain a millisecond timestamp.

To achieve this, we need to add the following methods to the Block:
1. `MilliTimestamp`: returns block timestamp in milliseconds level.
2. `SecondsTimestamp`: returns block timestamp in seconds level.
3. `NextMilliTimestamp`: returns next block timestamp in milliseconds level, if stored millisecond, add 500ms, otherwise 1s. 
4. `NextSecondsTimestamp`: returns next block timestamp in milliseconds level,  if stored millisecond, add 500ms, otherwise 1s, then keep seconds part.

**In op-geth, all previous internal direct accesses to the block timestamp(`block.Time()` or `header.Timestamp`) will be replaced according to the actual situation with the above four methods.**

The benefit of this approach is that it necessitates distinguishing whether each access to the timestamp is in milliseconds or seconds, forcing developers to clarify when to use seconds and when to use milliseconds. This helps avoid bugs caused by ambiguity. Additionally, the old methods are removed, ensuring that all modifications are accounted for without omissions.

### Implement

- [x] Step 1: Add `MilliTimestamp`, `SecondsTimestamp`, `NextMilliTimestamp`, and `NextSecondsTimestamp` for block/header/payload structs. And change block/header/payload `Timestamp` to 'TempTimestamp', for debugging where to change the new method. 
- [x] Step 2: Change the second-level timestamp to the millisecond-level timestamp in miner and EngineAPI.
- [x] Step3: Fix UT errors.
- [x] Step4: Change block/header/payload `TempTimestamp` to 'Timestamp'.


beacon/engine/types.go
```diff
@@ -20,12 +20,19 @@ import (
...
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/trie"
 )
 
+var (
+	OldBlockMillisecondsInterval uint64 = 1000
+	NewBlockMillisecondsInterval uint64 = 500
```
flywukong, line 33: It is best to add comments to the hard coded  which may change in the future. It is better to add a config to set this variable later to make to changable.

RenRick (approved):

flywukong (approved):
LGTM

will-2012 (approved):

core/types/block.go
```diff
@@ -101,14 +107,33 @@ type headerMarshaling struct {
...
+		return 0
+	}
+	return uint256.NewInt(0).SetBytes2(h.MixDigest[:2]).Uint64()
+}
+
+func (h *Header) MilliTimestamp() uint64 { return h.Time*1000 + h.millisecondes() }
+
+func (h *Header) NextMilliTimestamp() uint64 {
+	if h.MixDigest == (common.Hash{}) {
+		return h.Time*1000 + OldBlockMillisecondsInterval
```
atenjin, line 130: hey why hardcode this, other rollup may do not use 1s as the blockinterval

PR #279, opened 2025-04-01
feat: parse block interval from engine api

joey0612:
### Description

The op-node writes the BlockInterval into the third byte of the `PrevRandao` in `PayloadAttributes`, with each unit representing 500ms.

### Rationale

op-geth has more control over block time. Previously, if the BlockTime configuration of op-node was greater than 1 second, the excess time would be wasted.

### Example

NO.

### Changes

Notable changes:
NO.

### DIFF
--- a/core/types/block.go
+++ b/core/types/block.go
@@ -133,7 +133,7 @@ func (h *Header) BlockMillisecondTimeUnit() uint64 {
 	if h.MixDigest == (common.Hash{}) {
 		return DefaultBlockIntervalUintCount
 	}
-	count := uint256.NewInt(0).SetBytes2(h.MixDigest[2:2]).Uint64()
+	count := uint256.NewInt(0).SetBytes1(h.MixDigest[2:3]).Uint64()
 	if count == 0 {
 		return DefaultBlockIntervalUintCount
 	}
--- a/core/types/block.go
+++ b/core/types/block.go
@@ -33,8 +33,8 @@ import (
 )
 
 var (
-	DefaultBlockIntervalUintCount uint64 = 2
-	BlockMillisecondsIntervalUint uint64 = 500
+	DefaultBlockIntervalUintCount uint64 = 4
+	BlockMillisecondsIntervalUint uint64 = 250
 )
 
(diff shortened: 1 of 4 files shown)


RenRick (dismissed):

will-2012 (dismissed):

core/types/block.go
```diff
@@ -126,10 +126,22 @@ func (h *Header) millisecondes() uint64 {
 func (h *Header) MilliTimestamp() uint64 { return h.Time*1000 + h.millisecondes() }
 
 func (h *Header) NextMilliTimestamp() uint64 {
+	return h.MilliTimestamp() + h.BlockMillisecondTime()
+}
+
+func (h *Header) BlockMillisecondTimeUnit() uint64 {
 	if h.MixDigest == (common.Hash{}) {
-		return h.Time*1000 + OldBlockMillisecondsInterval
+		return DefaultBlockIntervalUintCount
```
flywukong, line 134:
```
   if h.MixDigest == (common.Hash{}) || len(h.MixDigest) < 3 {
        return DefaultBlockIntervalUintCount
    } 
```   maybe better

File: miner/worker.go [hunk 2]
```diff
@@ -1151,6 +1151,21 @@ func (g *generateParams) millisecondes() uint64 {
 
 func (g *generateParams) MilliTimestamp() uint64 { return g.timestamp*1000 + g.millisecondes() }
 
+func (g *generateParams) BlockMillisecondTimeUnit() uint64 {
+	if g.random == (common.Hash{}) {
```
flywukong, line 1155: if h.MixDigest == (common.Hash{}) || len(h.MixDigest) < 3

core/types/block.go
```diff
@@ -33,8 +33,8 @@ import (
 )
 
 var (
-	OldBlockMillisecondsInterval uint64 = 1000
-	NewBlockMillisecondsInterval uint64 = 500
+	DefaultBlockIntervalUintCount uint64 = 2
```
flywukong: DefaultBlockIntervalUintCount may need change to 4 if the BlockMillisecondsIntervalUint change to 250ms

flywukong (approved):

PR #282, opened 2025-04-18
feat: support config proposer block multiplier in nodebufferlist

sysvm:
### Description

Support config proposer block multiplier in nodebufferlist which can provide more withdrawal proof.

### Rationale

N/A

### Example

N/A

### Changes

Notable changes:
* Add `--pathdb.proposeblockmultiplier` flag in op-geth.


cmd/utils/flags.go
```diff
@@ -340,6 +340,12 @@ var (
 		Value:    pathdb.DefaultProposeBlockInterval,
 		Category: flags.StateCategory,
 	}
+	PathDBProposeBlockMultiplierFlag = &cli.Uint64Flag{
+		Name:     "pathdb.proposeblockmultiplier",
+		Usage:    "Multiplier for proof submission interval (e.g. 2 means providing proofs of 2*proposeblock)",
+		Value:    0,
```
joey0612, line 346: Default is 1 or 2?

joey0612 (approved):

cmd/utils/flags.go
sysvm: 0 means no need to config.

will-2012 (approved):

PR #286, opened 2025-05-13
feat: support generate witness in txdag async process;

galaio:
### Description

This PR will integrate stateless into TxDAG generation, it aims to optimize the performance of witness generation, and merge rw record, witness collection with TxDAG.

This PR will be based on the PR https://github.com/bnb-chain/op-geth/pull/287.

When `--parallel.txdagwitnessgen` is enabled, read-only states will no longer be submitted to the trie prefetcher. Instead, txdag will be queried asynchronously, but the original parallel processing logic will still be retained for states reading caused by writes.

### Example

Add these parameters to geth:
```bash
--selfstateless.validate --parallel.txdag --parallel.txdagwitnessgen
```
`--parallel.txdag` will enable the TxDAG generation feature, while parallel.txdagsenderpriv specifies the private key of the sender for the transaction into which the generated TxDAG data is inserted.

### Changes

Notable changes:
* feat: support generate witness in txdag async process;
* ...

### DIFF
--- a/core/state/mvstates.go
+++ b/core/state/mvstates.go
@@ -609,14 +609,6 @@ func (s *MVStates) handleRWEvents(items []RWEventItem) {
 		// recorde current as cannot gas fee delay
 		case CannotGasFeeDelayRWEvent:
 			s.asyncRWSet.cannotGasFeeDelay = true
-		// handle witness generation events
-		case ExecutionDoneEvent, AccReadFromDBEvent, SlotReadFromDBEvent:
-			log.Debug("witness generation event", "event", item.Event, "asyncWitnessRunning", s.asyncWitnessRunning, "trieDB", s.trieDB == nil)
-			// if no trieDB, skip witness generation
-			if !s.asyncWitnessRunning || s.trieDB == nil {
-				continue
-			}
-			s.witnessEventCh <- item
 		}
 	}
 	// handle last tx rw set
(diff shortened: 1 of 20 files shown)


core/state/mvstates.go
```diff
@@ -454,6 +506,49 @@ func (s *MVStates) asyncRWEventLoop() {
...
+	defer s.asyncWG.Done()
+	for {
+		select {
+		case item, ok := <-s.witnessEventCh:
+			if !ok {
+				return
+			}
+			switch item.Event {
+			case OriginAccReadEvent:
+				key := common.Hash{}
```
flywukong, line 523: key should be item.Addr?

core/state/mvstates.go
```diff
@@ -454,6 +506,49 @@ func (s *MVStates) asyncRWEventLoop() {
...
+				if s.trieCache[key] == nil {
+					trie, err := s.trieDB.OpenTrie(item.StateRoot)
```
flywukong, line 525: we should set the item.StateRoot in the event item

core/state/mvstates.go
```diff
@@ -454,6 +506,49 @@ func (s *MVStates) asyncRWEventLoop() {
...
+						continue
+					}
+					s.trieCache[key] = trie
+				}
+				// access account in trie
+				s.trieCache[key].GetAccount(item.Addr)
+			case OriginSlotReadEvent:
+				key := crypto.Keccak256Hash(item.Addr.Bytes())
+				if s.trieCache[key] == nil {
+					trie, err := s.trieDB.OpenStorageTrie(item.StateRoot, item.Addr, item.StorageRoot, nil)
```
flywukong, line 538: we should set the item.StorageRoot in the event item

core/state/mvstates.go
```diff
@@ -695,6 +797,55 @@ func (s *MVStates) RecordCannotDelayGasFee() {
 	s.rwEventCacheIndex++
 }
 
+func (s *MVStates) RecordOriginAccRead(addr common.Address) {
+	if !s.asyncRunning || !s.recordingRead {
+		return
+	}
+	if s.rwEventCacheIndex < len(s.rwEventCache) {
+		s.rwEventCache[s.rwEventCacheIndex].Event = OriginAccReadEvent
+		s.rwEventCache[s.rwEventCacheIndex].Addr = addr
```
flywukong: may need to set the root info here

